PR #3, opened 2021-06-17
Improve user ID regexp

defaultxr:
This change makes more clear the elements parsed from the user ID input. It now accepts `@` at the start, and allows the user to specify their homeserver after a `:`, defaulting to `matrix.org` if unspecified (as was the previous behavior).

### DIFF
--- a/ement.el
+++ b/ement.el
@@ -126,13 +126,14 @@ session and log in again."
                                  (ement-alist 'username (read-string "User ID: ")
                                               'password (read-passwd "Password: ")))))
                  (list username password token transaction-id)))
-  (pcase-let* ((hostname (if (string-match (rx ":" (group (1+ anything))) user-id)
-                             (match-string 1 user-id)
-                           "matrix.org"))
+  (string-match (rx (? "@") (group (1+ (not ":"))) (? ":") (group (* anything))) user-id)
+  (pcase-let* ((hostname (if (string= "" (match-string 2 user-id))
+                             "matrix.org"
+                           (match-string 2 user-id)))
                ;; FIXME: Lookup hostname from user ID with DNS.
                ;; FIXME: Dynamic port.
                (server (make-ement-server :hostname hostname :port 443))
-               (user (make-ement-user :id user-id))
+               (user (make-ement-user :id (match-string 1 user-id)))
                (transaction-id (or transaction-id (random 100000)))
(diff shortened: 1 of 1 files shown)


alphapapa:
Thanks, this looks like a good idea.

Rather than calling `string-match` directly, we could use `pcase-let*`'s ability to match regexps and bind groups to variables.  Would you mind making that change?

defaultxr:
I am not familiar with `pcase-let*` to be honest (if that wasn't already obvious), but I'll take a look into it and see about making that change as it's probably cleaner than how I wrote it.

alphapapa:
Thanks.  By the way, you should almost always make PRs from a feature branch on your fork rather than from your master branch.

defaultxr:
To be honest I'm not sure how to do what you're asking with `pcase-let*` and haven't been able to find any examples. I've never used `pcase` before and the documentation doesn't really make it clear how to bind regexp groups to variables; I presume it's possible to do something like `(pcase-let* (((user host) (rx REGEXP...) STRING)) (list user host))` to bind user and host to groups in the `REGEXP` form but am not really figuring it out.

alphapapa:
`pcase` is a powerful tool.  Its documentation isn't always easy to understand, but do see the Elisp manual section about it, which goes into more detail than the docstring and gives examples.  The docstring does describe how to use the `rx` matcher with `let` to bind match groups, but in the end, learning how to do something with `pcase` sometimes requires a bit of trial-and-error.  If you don't want to dive any deeper into pcaseland, I'll make the changes myself sometime.

alphapapa:
Actually, I was mistaken in that only `pcase` seems to support binding `rx` groups, not `pcase-let` (though this might be a bug in Emacs, and it might be fixed in newer versions).  For example:
```el
(pcase-let* ((user-id "@alphapapa:matrix.org")
             ((rx bos (optional "@")
                  (let username (1+ (not (any ":")))) ":"
                  (let hostname (optional (1+ (not (any blank))))))
              user-id))
  (list username hostname))
;;=> (nil nil)

(pcase "@alphapapa:matrix.org"
  ((rx bos (optional "@")
       (let username (1+ (not (any ":")))) ":"
       (let hostname (optional (1+ (not (any blank))))))
   (list username hostname)))
  ;;=> ("alphapapa" "matrix.org")
```
So I had to use `string-match` in the end.  I also made a few other improvements at the same time.  Thanks for reporting this.

PR #5, opened 2021-07-26
Silence byte-compiler

akater:


alphapapa:
I don't have a byte-compiler warning related to that feature.  The file compiles cleanly on my system.  What version of Emacs are you using, installed from where?  Are you using the sandbox configuration or your personal one?

akater:
Emacs 27.2

It happens on Emacsen built `--without-x`.  Namely:

ement is compiled with

`emacs -batch -q --no-site-file -L . -f batch-byte-compile *.el`

and
`emacs --batch -q --no-site-file --eval "(message \"%s\" (fboundp 'mwheel-scroll))"` outputs nil.

I just installed Emacs 26 with X support, and
`emacs-26 --batch -q --no-site-file --eval "(message \"%s\" (fboundp 'mwheel-scroll))"`
`t`

then reinstalled it without X support, and
`emacs-26 --batch -q --no-site-file --eval "(message \"%s\" (fboundp 'mwheel-scroll))"`
`nil`

Emacsen built `--without-x` likely should not ship `mwheel.el` in the first place (when I try it, I get “Bad binding” error) but it so happens that they do, so require'ing mwheel actually better be replaced with compile-time check for X support but I still don't know how to do that (if it's possible at all).  Anyway, even if upstream addresses the issue by excluding mwheel, error message for missing `mwheel` library will be more informative than a warning about unfbound `mwheel-scroll`, and will be catched earlier.

alphapapa:
Thanks.  I don't have any Emacs built without X support.

> Emacsen built `--without-x` likely should not ship `mwheel.el` in the first place

AFAIK some text terminals can send mouse events, so it might be useful without X.

Anyway, thanks for reporting.  I'll fix this soon.

PR #6, opened 2021-07-27
Search for password with auth-source

akater:
…and reject incorrect user-ids earlier.

This changes the reading interface a little; see the bit with `minibuffer-message`.

We're looking for `:user` and `:host` as they appear in the ID.  I registered my matrix account via an identity server so I actually recorded the password under that hostname and symlinked the `akater@matrix.org` entry to it.  I'm not sure how to deal with this properly.

We do not try to record passwords and while this convenience feature may be added, I'm certain it should be disabled by default.

### DIFF
--- a/ement.el
+++ b/ement.el
@@ -129,15 +129,15 @@ Writes the session file when Emacs is killed."
 Run with one argument, the session synced."
   :type 'hook)
 
+(defvar ement-user-id-history nil
+  "Default variable to store user-id history in.")
 
-
-
-
-
-
-
-
-
+(defcustom ement-keep-user-id-history nil
+  "When set to a variable, keep in it user-ids submitted to ement-connect."
+  :type '(choice (const nil)
+                 (variable-item ement-user-id-history)
+                 variable)
+  :group 'ement)
 
 ;;;; Commands
 
@@ -164,13 +164,10 @@ the port, e.g.
                    ;; Session available: use it.
                    (list :session (ement--read-session))
                  ;; Read username and password.
-                 (list :user-id (read-string "User ID: ")
-                       :password (read-passwd "Password: "))))
+                 (ement--read-user-credentials nil
+                                               ement-keep-user-id-history)))
   (cl-labels ((new-session ()
-                           (unless (string-match (rx bos "@" (group (1+ (not (any ":")))) ; Username
-                                                     ":" (group (optional (1+ (not (any blank)))))) ; Server name
-                                                 user-id)
-                             (user-error "Invalid user ID format: use @USERNAME:SERVER"))
+                           (ement--parse-user-id user-id)
                            (let* ((username (match-string 1 user-id))
                                   (server-name (match-string 2 user-id))
                                   (uri-prefix (or uri-prefix (ement--hostname-uri server-name)))
@@ -753,6 +750,57 @@ and `session' to the session.  Adds function to
   (pcase-let* (((cl-struct ement-event (content (map topic))) event))
     (when topic
       (setf (ement-room-topic room) topic))))
+
+(defun ement--parse-user-id (user-id)
+  "Alters match-data."
+  (string-match
+   (rx bos "@" (group (1+ (not (any ":"))))           ; Username
+       ":" (group (optional (1+ (not (any blank)))))) ; Server name
+   user-id))
+
+(defun ement--read-username-and-server (prompt &optional initial-input history
+                                               default-value
+                                               inherit-input-method)
+  "Read Matrix user ID, submitted in the form \"@username:server\".
+
+Return (user-input . plist) where plist is `auth-source-search'-compatible plist."
+  (cl-loop
+   as user-id = (read-string prompt
+                             initial-input history default-value
+                             inherit-input-method)
+   if (not (ement--parse-user-id user-id))
+   do (minibuffer-message "Invalid user ID format: use @USERNAME:SERVER")
+   else return (cons user-id
+                     (list :user (match-string 1 user-id)
+                           :host (match-string 2 user-id)))))
+
+(defun ement--read-user-credentials (&optional initial-input history
+                                               default-value
+                                               inherit-input-method)
+  "Read Matrix user credentials.  Return plist (&key user-id password).
+
+Optional arguments refer to reading user ID."
+  (cl-destructuring-bind (user-id . plist) (ement--read-username-and-server
+                                            "User ID: "
+                                            initial-input history
+                                            default-value
+                                            inherit-input-method)
+    (list :user-id user-id
+          :password
+          (save-match-data
+            ;; save-match-data is necessary at least when
+            ;; auth-source does not use cache
+            (or (when-let ((auth-source-secret
+                            (when (fboundp 'auth-source-search)
+                              (plist-get (car (apply #'auth-source-search
+                                                     :max 1
+                                                     plist))
+                                         :secret))))
+                  (if (functionp auth-source-secret)
+                      (funcall auth-source-secret)
+                    auth-source-secret))
+                (read-passwd "Password: "))))))
+
 
 ;;;; Footer
 
--- a/ement.el
+++ b/ement.el
@@ -93,6 +93,8 @@ by users; ones who do so should know what they're doing.")
           (timeline (lazy_load_members . t))))
   "Default filter for sync requests.")
 
+
+
 ;; From other files.
 (defvar ement-room-avatar-max-width)
 (defvar ement-room-avatar-max-height)
@@ -126,6 +128,16 @@ Writes the session file when Emacs is killed."
   "Hook run after initial sync.
 Run with one argument, the session synced."
   :type 'hook)
+
+
+
+
+
+
+
+
+
+
 
 ;;;; Commands
 


alphapapa:
Thanks.  I'll make some comments as a review.  In the meantime, I just pushed a new feature, so this branch needs to be rebased on current master.

ement.el
```diff
@@ -141,10 +139,6 @@ session and log in again."
         (let ((ement-save-session nil))
           (call-interactively #'ement-connect)))
     ;; Log in to new session.
-    (unless (string-match (rx bos "@" (group (1+ (not (any ":")))) ; Username
-                              ":" (group (optional (1+ (not (any blank)))))) ; Server name
-                          user-id)
-      (user-error "Invalid user ID format: use @USERNAME:SERVER"))
     (let* ((username (match-string 1 user-id))
```
alphapapa, line 142: Since you've removed the `string-match` and are returning a plist instead, you need to bind the variables to the returned plist values, no?  Why rely on the global match data across function calls when you're doing that in a separate function?  (Please use `pcase-let*`).

ement.el
```diff
@@ -222,6 +216,51 @@ call `pop-to-buffer'."
 
 ;;;; Functions
 
+(defun ement--read-username-and-server (prompt &optional initial-input history
+                                               default-value
+                                               inherit-input-method)
+  "Read Matrix user ID, submitted in the form \"@username:server\".
+
+Return (user-input . plist) where plist is `auth-source-search'-compatible plist."
+  (cl-loop
```
alphapapa, line 225: I understand why you use the loop, but I think it's not necessary here.  On the rare occasion that a user mistypes the ID, they can call the command again.

ement.el
```diff
@@ -222,6 +216,51 @@ call `pop-to-buffer'."
 
 ;;;; Functions
 
+(defun ement--read-username-and-server (prompt &optional initial-input history
```
alphapapa, line 219: Do we really need all these extra arguments?  We aren't passing them, are we?

ement.el
```diff
@@ -222,6 +216,51 @@ call `pop-to-buffer'."
...
+   if (not (string-match
+            (rx bos "@" (group (1+ (not (any ":")))) ; Username
+                ":" (group (optional (1+ (not (any blank)))))) ; Server name
+            user-id))
+   do (minibuffer-message "Invalid user ID format: use @USERNAME:SERVER")
+   else return (cons user-id
+                     (list :user (match-string 1 user-id)
+                           :host (match-string 2 user-id)))))
+
+(defun ement--read-user-credentials (&optional initial-input history
```
alphapapa, line 238: Can we combine these two functions?  They seem to be very similar.

ement.el
```diff
@@ -222,6 +216,51 @@ call `pop-to-buffer'."
...
+                                               default-value
+                                               inherit-input-method)
+  "Read Matrix user credentials.  Return list (user-id password).
+
+Optional arguments refer to reading user ID."
+  (cl-destructuring-bind (user-id . plist) (ement--read-username-and-server
```
alphapapa, line 244: Please prefer `pcase` over `cl-destructuring-bind` (although if this function is eliminated, it won't matter).

ement.el
```diff
@@ -222,6 +216,51 @@ call `pop-to-buffer'."
...
+                                            "User ID: "
+                                            initial-input history
+                                            default-value
+                                            inherit-input-method)
+    (list user-id
+          (save-match-data
```
alphapapa, line 250: Generally, I think having to use `save-match-data` is a "code smell."  We shouldn't be relying on the match data across function calls.  (Again, this probably won't matter if we combine these functions.)

ement.el
```diff
@@ -222,6 +216,51 @@ call `pop-to-buffer'."
...
+            ;; save-match-data is necessary at least when
+            ;; auth-source does not use cache
+            (or (when-let ((auth-source-secret
+                            (when (fboundp 'auth-source-search)
```
alphapapa, line 254: I don't use auth-source, but is this `fboundp` check really necessary?  Having to use this might indicate that we should be doing something else.

alphapapa (changes requested):
Please also use a commit message in the same format as I use in this repo.  Thanks.

ement.el
akater: I'd gladly bind the parsed values directly but that requires the interface of `ement-connect` to change, e.g. its arglist would become `(&key user host password)` or just `(&rest properties)`.  Which one you'd prefer?

ement.el
akater: • I miss a good input history already and was going to add it.
• In general I tend to do this when it's clear how interfaces map to each other.  All standard read- functions have the same interface. I realise some will object to this.  You do; OK.  Unless you're not convinced that your read- functions should share this interface, I'll retract the unused args.

ement.el
akater: Sure.  It's just the way they were written, they looked better split.

ement.el
akater: Yes.  I hope you're fine with `ement-connect` supporting more rich input.

ement.el
akater: I'll drop it.

I tend to write code that would be easy to make optional.  Passwords is a matter delicate enough for this.  It has become a habit but it better be left in my libraries.

ement.el
alphapapa: I'd prefer to keep the interface as it is.  Is it really necessary to return a plist here?

ement.el
alphapapa: I don't think it's necessary for these prompts to have their own input histories, is it?  Most users will type in their MXID and password no more than once per session, and less often if they save the session.  I'd prefer to keep everything as simple as possible.

ement.el
alphapapa: If we're `require`ing `auth-source`, then the function should be available, right?  So it shouldn't be necessary to test this.

Supporting things optionally in Emacs is usually a bit of a hassle.  I'd prefer to keep the code as simple as possible.

ement.el
akater: We're reading both user-id and password in the `interactive` form.  user-id needs to be split to do a password lookup.  As we read password in the `interactive` form, it has to be split in the `interactive` form.  Then it has to get split the same way the second time in the function body because bindings established within `interactive` form are not avaliable in the body (unless we use a let over defun).  I remember trying to capture the results of in-`interactive` computations once.  `&aux` didn't help, so let over defun was the only way to go.  Or we could simply reuse the saved match-data which I did initially.

Otherwise, we could simply repeat the splitting proceduce in the function body.  This means repeating the same computation twice.  If you're fine with that, I'll do it.  I myself would never do it if another solution would be readily available: if for any reason two results diverge, it will be a baffling bug, and any side-effects in the splitting procedure, or in other future in-`interactive` computations would become unsafe.

alphapapa: I think it's more important to keep the command's signature sensible (from a user's point of view) than to avoid calling `string-match` one extra time in a function that is called once per session.  I don't think the MXID-splitting regexp is going to change of its own accord, but it could go into a variable, or the splitting itself could be done in a function.

akater: OK, I'll re-split in the `let*` preserving the function signature.  Will re-submit tomorrow, likely.  I think I understand the way you want it good enough.

alphapapa:
By the way, if using `auth-source` presents any problems at all, I'm strongly inclined to not support it at all.  A few years ago I tried to support it in `matrix-client`, and I ended up writing this instead: https://www.reddit.com/r/emacs/comments/8lvda6/is_authsource_from_the_dark_side/

alphapapa:
Frankly, reading my old post there, I'm disinclined to support it at all.  Unless the problems I described there have been solved, it seems like it will inevitably present problems for end users, who will come here and complain when their old password is used unconditionally.  Ted said there that he was going to work on it, but I haven't checked to see if anything's changed since then.

And if we are going to use `auth-source`, we probably need to use it to store tokens as well, because Matrix isn't intended for clients to log in with a username and password every time (doing that creates a new session every time, which end up in very long lists of old sessions in Element settings).

So, not to dismiss your work on this, but I think we should consider carefully if this is even a good idea.

akater:
Well, this is why I have a habit of making such stuff optional from the start.  The quality of handling passwords in Emacs is below average.

I could write and publish a standalone `ement-auth-source` package if there's an `ement-read-credentials` function with an established interface that could be cleanly advised or redefined (or maybe `ement-credentials-reader` variable?) to hook to auth-source.   All the hairy password issues would live in that standalone repository, and the feature would be disabled by default.

Or it could be a part of ement but not loaded by default.

My efforts won't be lost; I'll be using password lookup anyway but I'd like to contribute as much of it as I can.

alphapapa:
> I could write and publish a standalone `ement-auth-source` package if there's an `ement-read-credentials` function with an established interface that could be cleanly advised or redefined (or maybe `ement-credentials-reader` variable?) to hook to auth-source. All the hairy password issues would live in that standalone repository, and the feature would be disabled by default.

Maybe I'm missing something, but couldn't you advise `ement-connect` to do that already?

> Or it could be a part of ement but not loaded by default.

I'd be more inclined to have it as a separate library that doesn't affect the existing code, yes.

> My efforts won't be lost; I'll be using password lookup anyway but I'd like to contribute as much of it as I can.

Sure, I don't mind making small changes to make it more flexible and extensible.

Thanks.

akater:
It does not look like advising interactive functions, like `ement-connect`, is possible this way.

Reading user-id is “a friendly interface”.  A user-friendly string is parsed into the data structure useful to multiple external agents: a -connect function, a password-retrieving function, proxy-related functions.  This structure better be accessible to all such.  However, right now everything happens in `ement-connect` which is user-facing and is thus relatively opaque.  I think our proxy-related discussions also suggest that this is an obstruction.

Otherwise, read-credentials has to be introduced and advised or replaced.  In fact, it is only the password part of it that should be advised, which roughly corresponds to my `ement--read-user-credentials`.  But in both cases, this would imply at least some code duplication which is a clear indication of a wrong design.  Code only wouldn't be duplicated if `ement-connect` (or a non-user-facing variation of it) accepted a plist or another comparatively rich data structure, and proxy arrangers / password lookups would use it as well.

Or I'd have to reimplement the entire `ement-connect`, duplicating all code from there and hope to get up to date with its updates.  At this point, I'd keep a branch where I patch `ement-connect` directly, keeping it for myself or those who follow my Gentoo repository, but not releasing it any wider.

alphapapa:
If it's really necessary, we can change `ement-connect` to take a plist.  I just want to preserve the ability to call it like `(ement-connect :id "@username:host" :password "PASSWORD")`.

akater:
If you're not convinced it's necessary, don't do it.

I would never design a -connect function, in any application, this way.  Certainly not when it's going to deal with a federated network.  But if our proxy- and auth-source- related incidents don't convince you, I think it will be counterproductive to attempt to redesign.  Maybe later.

There should likely be a separation into a user-facing high-level function `ement` and a lower-level `ement-connect`.  I made a write-up on what would in my opinion be a reasonable interface for `ement-connect` and (hypothetical high-level) `ement`.  According to it, `(ement-connect :id "@user:matrix.org" :password "***")` works, basically exactly as it does now.  If you think redesign is worth it, I'll post my vision of the interface as a separate issue.  I'm ready to work on implementation, too.  Lots of it is already in ement, only not decoupled.

alphapapa:
Maybe I'm missing something, but I don't understand why this would need to be so complicated.  It's just a command that reads an MXID and password and takes the first step in the chain of connecting and authenticating.  If it would be helpful to use keyword arguments in the signature, that's a very minor change.  

As I explained, I'm disinclined to support `auth-source` because of its API and bugs.  And as I explained in the chat room, I'm reluctant to add proxy support, because it's a complicated issue, very prone to mistakes and bugs, and something that I think is best handled at a higher layer or with a wrapper.

My intent for this client is to keep it simple, so I want to avoid complex designs and APIs.  However, I also want it to be flexible and extensible, so if there's room for improvement, feel free to make suggestions.  You should expect me to push back on anything that seems complex or unnecessary to me.

alphapapa:
By the way, since `ement-connect` accepts an MXID and password as arguments, it seems to me that you could easily write your own function to retrieve your MXID and password from `auth-source` or anything else, and then call `ement-connect` with them.  So I don't see why we need to change any of the signatures or APIs for it.  Am I missing something?

akater:
Let's see how keepass lookup gets implemented.  I'll adjust to that.

alphapapa:
See also https://github.com/alphapapa/ement.el/commit/2607ec3145d0172db071bf31068fd980a19013c1

PR #17, opened 2021-08-09
Docs: Describe Pantalaimon usage

akater:
Port in the example is 8009.  This is the default port according to
man pantalaimon(5) as provided by pantalaimon-0.12.5. The manpage is
dated 2019-08-05.

I tested the suggested code as is, it works.

### DIFF
--- a/README.org
+++ b/README.org
@@ -174,6 +174,16 @@ Emacs may not display certain symbols and emojis well by default.  Based on [[ht
   (set-fontset-font t 'unicode "Noto Emoji" nil 'append)
 #+END_SRC
 
+*** E2E support with Pantalaimon
+Pantalaimon users may connect with
+
+#+BEGIN_SRC elisp
+(ement-connect :user-id "@USERNAME:SERVER" :password "PASSWORD"
+               :uri-prefix "http://localhost:8009")
+#+END_SRC
+
+where 8009 is to be replaced with the port specified in your Pantalaimon configuration.
+
 * Rationale
 
 Why write a new Emacs Matrix client when there is already [[https://github.com/alphapapa/matrix-client.el][matrix-client.el]], by the same author, no less?  A few reasons:


alphapapa:
Thanks, this will be good to have in the readme.

I think it would be best if the password were not necessarily passed in the function call, rather it could be read from the user.  That should only take a few changes in `ement-connect`, and it would allow users to not have to write their password in plain-text in a file somewhere.  What do you think?

akater:
Sure.  I already told, annoyingly enough, about the design I consider natural: a read-connection-properties function and a bunch of functions bringing the corresponding object (an arglist, as of now) to a canonical form.  It covers this as well.

akater:
Note however that pantalaimon won't run without any config at all.  I thought whether this should be reflected in the README but nothing came to mind.  Here's more or less what I have for Ement's Gentoo package:

```org
#+title: Quick start with E2E encryption in Ement.el

* User config

You may
#+begin_src sh
ln -s /etc/ement/pantalaimon.conf ~/.config/pantalaimon/pantalaimon.conf
#+end_src

or do something equivalent.

* Starting the server

Start the server with e.g.
#+begin_src sh
pantalaimon --config ~/.config/pantalaimon/pantalaimon.conf
#+end_src

* Starting Ement with Pantalaimon

Follow Ement's instructions.
```

The last sentence is going to become a link to the corresponding section in Ement's README.

alphapapa:
> Sure. I already told, annoyingly enough, about the design I consider natural: a read-connection-properties function and a bunch of functions bringing the corresponding object (an arglist, as of now) to a canonical form. It covers this as well.

You'll have to forgive me, but that doesn't seem natural to me, and "a bunch of functions" don't seem necessary to me.  Making the necessary change to `ement-connect` to handle the case when a password isn't passed should be only a few lines.  Let me see...

alphapapa:
There, that works.

alphapapa:
> Note however that pantalaimon won't run without any config at all. I thought whether this should be reflected in the README but nothing came to mind. Here's more or less what I have for Ement's Gentoo package:

Yeah, I don't want to explain how to configure other software in this software's readme.  I think it will be enough to say something like, "After configuring Pantalaimon according to its documentation..."

PR #18, opened 2021-08-11
Fix: (ement-room--room-display-name) check for empty room first

treed:
The spec for calculating room names is, I think, confusingly specified.

I believe it should not be considered sequentially because subparagraph iii can't be reached if you consider i first.

This change re-orders the branches so that it will check for rooms where the total member count is <= 0 before considering the other possibilities.

### DIFF
--- a/ement-room.el
+++ b/ement-room.el
@@ -1590,16 +1590,16 @@ data slot."
                                 summary))
                     ;; TODO: Disambiguate hero display names.
                     (when hero-ids
-                      (cond ((>= (length hero-ids) (1- (+ joined-count invited-count)))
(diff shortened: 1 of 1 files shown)


alphapapa:
For future reference, see conversation around https://matrix.to/#/!NicAJNwJawmHrEhqZs:matrix.org/$ZRsQAiOncpVLkDl2F2EO0ySauDfvfFk3fHmwm9Kf2U4?via=matrix.org&via=t2bot.io&via=hpdeifel.de (someday we should make those links usable in Ement...)

alphapapa:
@treed Thanks, and thanks for your patience.

treed:
Yeah, no worries. Glad to see it fixed!

PR #23, opened 2021-08-12
Correct straight instructions

gi-yt:


alphapapa:
Thanks, but I decided to remove the Straight instructions since I don't use it myself.

PR #26, opened 2021-08-14
Fix: (ement-room-compose-from-minibuffer) Inherit minibuffer's input method properly

### LINKED ISSUE #16: Compose buffer does not inherit the room's input method.
The compose buffer does not actually inherit the room's input method.
To reproduce this, do the following:

1. Visit a room in ement.
2. Set the room's input method to tamil-itrans (C-x C-m C-\\
   tamil-itrans RET).
3. Press RET and type something.
4. Press C-c ' to switch to the compose buffer.
5. Notice how the compose buffer has NOT inherited the input method
   i.e., its input method is rfc1345 (U in modeline) instead of tamil-itrans.
6. Press C-x C-s to quit the compose buffer.
7. Press C-g to cancel the message.
8. The room's input method has changed to rfc1345.  This is unexpected
   as input methods are supposed to be buffer local.
9. Further invocation of _any_ minibuffer command makes it use the
   tamil-itrans method.  This should NOT happen.  M-x should not inherit
   the input method but it does.  To actually use it, you need to press C-\\
   to switch back to the default input method.

In essence, the input methods go out of sync.

Now, if you change the compose buffer's input method.  A different
behaviour is observed.  To reproduce:

1. Visit a room in ement.
2. Set the room's input method to tamil-itrans (C-x C-m C-\
   tamil-itrans RET).
3. Press RET and type something.
4. Press C-c ' to switch to the compose buffer.
5. Notice how the compose buffer has NOT inherited the input method
   i.e., its input method is rfc1345 instead of tamil-itrans.
6. Change the input method to TeX (C-x C-m C-\ C-\ TeX RET).
7. Press C-x C-s to quit the compose buffer.
8. Press C-g to cancel the message.
9. The room's input method has changed to TeX.  This is, again,
   unexpected as input methods are supposed to be buffer local.
10. Further invocation of send message makes it use the TeX method.
This is different from previous cases' 10 since now M-x does not use the
TeX method (this is the expected behaviour).

This is again unexpected since changing the input method of the
compose buffer should not change the room's input method.  There's an
extra C-\ in step 6 when changing the input method to disable the
tamil-itrans method.  Somehow switching to the compose buffer makes
EVERY minibuffer command use the input method.  

EDIT: I'll try to fix this tomorrow, as it is quite late right now.

vizs:
Fixes #16.

* `pop-to-buffer' seems to deactivate the buffer's input method.
This meant that users had to explicitly switch to the correct input
method by pressing C-\.

* Without explicitly disabling the input method of the minibuffer,
every minibuffer command inherited the room's input method.  This is
not the desired behaviour.  The solution seems to be to explicitly
deactivate the minibuffer's input method before calling
`abort-recursive-edit'.

### DIFF
--- a/ement-room.el
+++ b/ement-room.el
@@ -2423,9 +2423,8 @@ To be called from a minibuffer opened from
                          (add-hook 'window-configuration-change-hook show-buffer-fn-symbol)))))
     (fset compose-fn-symbol compose-fn)
     (add-hook 'minibuffer-exit-hook compose-fn-symbol)
-    ;; If minibuffer's input method is not deactivated, then all
-    ;; minibuffer commands inherit the room's input method.  This is
-    ;; not the correct behaviour.
+    ;; Deactivate minibuffer's input method, otherwise subsequent
+    ;; minibuffers will have it, too.
     (deactivate-input-method)
     (abort-recursive-edit)))
 
--- a/ement-room.el
+++ b/ement-room.el
@@ -2370,12 +2370,9 @@ use it as the initial message contents."
                  (cl-assert ement-room) (cl-assert ement-session)
                  (list ement-room ement-session)))
   (let* ((compose-buffer (generate-new-buffer (format "*Ement compose: %s*" (ement-room--room-display-name ement-room))))
-         (input-method current-input-method)
          (send-message-filter ement-room-send-message-filter))
     (with-current-buffer compose-buffer
       (ement-room-init-compose-buffer room session)
-      (when input-method
-        (set-input-method input-method))
       (setf ement-room-send-message-filter send-message-filter)
       ;; TODO: Make mode configurable.
       (when body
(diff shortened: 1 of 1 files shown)


ement-room.el
```diff
@@ -2415,17 +2412,21 @@ To be called from a minibuffer opened from
...
+                                                (pop-to-buffer compose-buffer)
+                                                (set-input-method input-method))))
                          (fset show-buffer-fn-symbol show-buffer-fn)
                          (add-hook 'window-configuration-change-hook show-buffer-fn-symbol)))))
     (fset compose-fn-symbol compose-fn)
     (add-hook 'minibuffer-exit-hook compose-fn-symbol)
+    ;; If minibuffer's input method is not deactivated, then all
+    ;; minibuffer commands inherit the room's input method.  This is
+    ;; not the correct behaviour.
+    (set-input-method nil)
```
alphapapa, line 2429: Good catch, thanks.  We should probably use `deactivate-input-method` though; according to Emacs docs, that's the right way.

alphapapa (changes requested):
Thanks.

ement-room.el
vizs: I will do that.

alphapapa:
Thanks for your work on this!  :)

PR #29, opened 2021-08-17
Add: support imenu in ement-room-mode

treed:
I cleaned it up and made a commit. I tried to guess how to name stuff and where it belongs based on existing content, but by all mean let me know if it needs further massaging.

### DIFF
--- a/ement-room.el
+++ b/ement-room.el
@@ -1474,6 +1474,7 @@ and erases the buffer."
         '(ement-room--complete-members-at-point ement-room--complete-rooms-at-point)
         left-margin-width ement-room-left-margin-width
         right-margin-width ement-room-right-margin-width
+        imenu-create-index-function #'ement-room--imenu-index-function
         ;; TODO: Use EWOC header/footer for, e.g. typing messages.
         ement-ewoc (ewoc-create #'ement-room--pp-thing)))
 (add-hook 'ement-room-mode-hook 'visual-line-mode)
@@ -1664,6 +1665,19 @@ data slot."
   "Return STRING with \"%\" escaped.
 Needed to display things in the header line."
   (replace-regexp-in-string (rx "%") "%%" string t t))
+
+(defun ement-room--generate-imenu-item (node)
+  "Generate an imenu cons cell from NODE."
+  (cons (format-time-string (string-trim ement-room-timestamp-header-with-date-format) (cadr (ewoc-data node))) (ewoc-location node)))
+
+(defun ement-room--timestamp-event-p (node)
+  "Return non-nil if NODE is a timestamp event."
+  (let ((event (ewoc-data node)))
+    (and (listp event) (eq (car event) 'ts))))
+
+(defun ement-room--imenu-index-function ()
+  "Generate imenu index for the current room."
+  (map 'array #'ement-room--generate-imenu-item (ement-room--ewoc-collect-nodes ement-ewoc #'ement-room--timestamp-event-p)))
 
 ;;;;; Events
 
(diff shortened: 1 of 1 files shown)


treed:
I also wasn't sure if we wanted a separate defcustom for the format of imenu items.

alphapapa:
Hm, you know, at first I thought that using Imenu to go to timestamps wouldn't be that useful, but on second thought, yeah, that could be really helpful to go back in a room's history.

> I also wasn't sure if we wanted a separate defcustom for the format of imenu items.

Hm, maybe.  What did you have in mind?

I'll make more specific comments a little later, a bit under the weather at the moment.  :)

alphapapa:
I took the liberty of making a few adjustments and rebasing on master.  Quick question, though: I noticed that Imenu seems to replace spaces in the formatted string with periods.  Interestingly, `counsel-imenu` renders them as spaces.  Do you have any idea what's going on here?

treed:
> Hm, you know, at first I thought that using Imenu to go to timestamps wouldn't be that useful, but on second thought, yeah, that could be really helpful to go back in a room's history.

Yeah, I wasn't sure either. I mostly did it as an exercise to learn how imenu support works. But using it as a way to kind of bounce around history turns out to be kind a nice. And we can always add more or different imenu support later, I suppose, if we think of anything else useful. (Although I guess straight up removing it might get complaints if people start getting used to it.)

> Hm, maybe. What did you have in mind?

Well, mostly to avoid having to trim the string every time, but also in case anyone wanted to have a different format there. I just kinda used the most appropriate existing format string.

> I took the liberty of making a few adjustments and rebasing on master. 

Yeah, that's fine. I was trying to keep up with rebases because my emacs is currently pointing at this branch.

> Quick question, though: I noticed that Imenu seems to replace spaces in the formatted string with periods. Interestingly, counsel-imenu renders them as spaces. Do you have any idea what's going on here?

I hadn't noticed that. I use vertico and they render there as spaces. Where do you see the periods? In a sandbox or something?

> Also, we might want to consider using Imenu more idiomatically by using hierarchical sections with each day having time-only timestamps within it. But I don't know if that would be worth it.

It might be worth it if we expect people to have these buffers open for weeks. If it's only 4-5 days then it might be fine as is?

alphapapa:
> I hadn't noticed that. I use vertico and they render there as spaces. Where do you see the periods? In a sandbox or something?

Yeah, a mostly basic Emacs config that I use for running Ement most of the time.  I don't think I've changed anything in it that could cause it, but I guess I'd have to do an actual `emacs -q` to be sure (or `./makem.sh -s interactive` to run in a clean sandbox).

> It might be worth it if we expect people to have these buffers open for weeks. If it's only 4-5 days then it might be fine as is?

Probably so, yeah.  We can worry about that if the time ever comes.  (I have tested loading like 4,000 old messages in a room, and it works pretty well, as long as you load them in smaller batches so the server has time to build the responses.)

treed:
In chat, @vizs points out `imenu-space-replacement`, and yes if I just use M-x imenu then I see the dots. Unsure what to do about this, if anything. Do people who use basic imenu become accustomed to that or do imenu-enabled modes avoid spaces?

vizs:
Well, I can only see prog-modes using imenu?  Org does have imenu setup and it does nothing special, it leaves the dot thing to the user.  I think following what org does is reasonable.  We can mention `imenu-space-replacement' in the docstring for confused users.

alphapapa:
I think we can just bind `imenu-space-replacement` to a space with `setq-local` in room buffers.

alphapapa:
Actually, I think we shouldn't do that, because Imenu does that by default for a good reason, as it explains in the docstring:

    Used when presenting the index in a completion buffer to make the
    names work as tokens.

So I guess this is ready to go!

PR #61, opened 2022-03-29
Fix malformed m.receipt events causing error in process sentinel

jfmcbrayer:
Sometimes I get this error: `error in process sentinel: Wrong type
argument: listp, "{\"ts\": 1553242544352}"` when syncing. It's
happening in the m.receipt event handler. It seems like for some
reason this event, but not all events, hasn't been decoded from json,
or needs to be decoded a second time.

Hopefully this patch is minimal and harmless. An alternative behavior
would be to drop the malformed receipt events.

### DIFF
--- a/ement.el
+++ b/ement.el
@@ -1179,7 +1179,10 @@ and `session' to the session.  Adds function to
                          ;; NOTE: The JSON map keys are converted to symbols by `json-read'.
                          ;; MAYBE: (Should we keep them that way?  It would use less memory, I guess.)
                          do (puthash (symbol-name user-id)
-                                     (cons (symbol-name event-id) (alist-get 'ts receipt))
+                                     (cons (symbol-name event-id)
+                                           (if (json-alist-p receipt )
+                                               (alist-get 'ts receipt)
+                                             (alist-get 'ts (json-parse-string receipt :object-type 'alist))))
                                      room-receipts)))))
 
 (ement-defevent "m.space.child"


alphapapa:
Thanks for the report and the patch, but this shouldn't be happening.  If it does, it indicates something else going on that needs to be understood first.  How often does this happen?  Does it happen in all rooms or only some?  

Can you find one of the malformed events to show me?  You could try using this as the handler to show a warning when one of them is processed:
```el
(ement-defevent "m.receipt"
  (ignore session)
  (pcase-let (((cl-struct ement-event content) event)
              ((cl-struct ement-room (receipts room-receipts)) room))
    (cl-loop for (event-id . receipts) in content
             do (cl-loop for (user-id . receipt) in (alist-get 'm.read receipts)
                         ;; Users may not have been "seen" yet, so although we'd
                         ;; prefer to key on the user struct, we key on the user ID.
                         ;; Same for events, unfortunately.
                         ;; NOTE: The JSON map keys are converted to symbols by `json-read'.
                         ;; MAYBE: (Should we keep them that way?  It would use less memory, I guess.)
                         do (condition-case nil
                                (puthash (symbol-name user-id)
                                         (cons (symbol-name event-id) (alist-get 'ts receipt))
                                         room-receipts)
                              (error (warn "Ement.el: Malformed m.receipt event?  %s" event)))))))
```

jfmcbrayer:
Yes, I'll try running with your debug handler to see if/when it happens again.

jfmcbrayer:
Yeah, I'm seeing *some* events containing these:

``` lisp
($155343163564388jTqWP:maunium.net (m.read (@xxxxxx:matrix.org . {"ts": 155312345263})))
```

(ids and names changed to protect the innocent.)

@xxxxxx:matrix.org could be either of two users, both on matrix.org.

If you would like the whole warning, I can send it by email maybe? Rather not post it here so as not to have to anonymize everyone in it.

alphapapa:
Thanks.

What I'll need to do next is make a way to log the raw JSON event so we can try to figure out why part of it isn't getting decoded.  I'll post it here when it's ready, probably in a branch, so you'll be able to use that to get the raw event.

AFAIK none of the metadata in the event should matter, i.e. the event ID, user ID, or timestamp.  What matters is the format and structure of the JSON.  So you can edit the metadata as desired.

In the meantime, I'll merge a version of that handler into `master` so this won't cause errors or warnings anymore.

alphapapa:
Ok, I pushed a branch to `wip/log-raw-events` (note: "I have not tested it, only proved it correct").  It logs every raw JSON response into a buffer named `*Ement API Log*`.  So when the warning pops up, you can switch to that buffer and search for the JSON containing that m.receipt event (probably by searching for the timestamp value, which will likely be unique in the log).  Then we should be able to compare it to other m.receipt events and figure out what's going on.

jfmcbrayer:
For some reason, I'm not getting the logs I would expect to be getting. I'm getting the usual warning on startup, but nothing is logged at that time. A bit later, the log looks like:

``` json
{}
{}
{}
{}
{}
{}
{}
{
  "event_id": "$RLRyxPGXoUT5sOY6t6pUIBrGxgZiXqPrwpYqEMrrs80"
}
{}
```

Which is a lot less than I'd expect. The logging code looks right to me, though, so presumably there's something else I'm doing wrong.

alphapapa:
I should have tested it first.  I neglected to change the right function call to make the log function take effect.  Please pull from that branch again and it should work now (I have tested it).

jfmcbrayer:
Okay, got a good log here. Trimming a bunch of stuff, we have this:

``` json
{
  "ephemeral": {
    "events": [
      {
        "type": "m.receipt",
        "content": {
          "$15532465765424XOTLF:matrix.org": {
            "m.read": {
              "@xxxxx:matrix.org": "{\"ts\": 1553453463352}"
            }
          },
          "$155331345634aWKe:matrix.xxxx.net": {
            "m.read": {
              "@xxxx:mainframe.xx": {
                "ts": 1574548595154
              }
            }
          },
          "$374894783215288jTqWP:xxxxxxx.net": {
            "m.read": {
              "@xxxxxxx:matrix.org": "{\"ts\": 15537364376473}"
            }
          }
        }
      }
    ]
  }
}
```

I'm getting this in the midst of all the other events on initial sync.

alphapapa:
@jfmcbrayer: Interesting: the last one is not even a valid timestamp, as it decodes to 2462-05-11 09:32:56 -0500 (but if the last digit is removed, it decodes to 2019-03-27 20:27:17 -0500, which is within the same year as the other two).

But that's not even the problem here, it being that the value of those user ID keys are strings containing JSON objects rather than the proper JSON maps directly.

So I can only guess that some clients have sent malformed m.receipt events, and that the Matrix server does not validate them.  I think this isn't the first example of this I've seen, although I don't recall specifics.

Given that, we need to be ready for malformed events of other types, so I've pushed this commit that demotes errors signaled while running event handlers to messages.  I'm guessing that these will be rare enough that it won't be worth writing workarounds for each type of malformed event like this (surely not for these years-old receipts, anyway), but if I'm wrong, we can add more specific workarounds in the future.

Please let me know if you find any more similar problems.  Thanks for reporting this.

PR #63, opened 2022-04-14
Add: WIP support for URL previews

rayes0:
Implements basic support for asynchronous URL previews, currently almost everything is working. Getting the preview, inserting the preview to the proper place, inserting later messages to the proper places, etc.

It can be turned on by the variable `ement-room-url-previews`, set to nil by default because some servers don't support URL previews (probably be best to mention something in the readme about this if it gets merged).

There are two things that currently don't work very well:
- Type errors when sender headers are enabled coming from somewhere in this block: https://github.com/alphapapa/ement.el/blob/4692961c642a236581e449465385c47310128613/ement-room.el#L2379
I'm not too sure why, all the macros make it hard to debug, though I suspect it is trivial. And everything appears to work fine except that it throws an error. The previews are inserted correctly, and even the headers are inserted correctly.
- Sometimes there is a race condition when the pretty printer is called in quick succession and the previews haven't had time to be inserted into the EWOC yet causing two previews to be inserted. I'm not sure how to solve this one, it appears how you did it with inserting retro messages was by setting a variable while loading and not starting a process it the variable is set. I'm not sure how that would work with URL previews, especially multiple URLs in the same message. Any help/suggestions would be appreciated :)

Feel free to edit, change, or delete anything you wish :)

### DIFF
--- a/ement-room.el
+++ b/ement-room.el
@@ -189,6 +189,10 @@ this one automatically.")
     '((t (:inverse-video t :extend t))))
   "Messages that mention the local user.")
 
+(defface ement-room-preview
+  '((t (:inherit ement-room-message-text)))
+  "Text of URL previews.")
+
 ;;;;; Options
 
 (defcustom ement-room-ellipsis "⋮"
@@ -206,6 +210,17 @@ this one automatically.")
 (defcustom ement-room-avatar-max-height 32
   "Maximum height in pixels of room avatars shown in header lines."
   :type 'integer)
+
+(defcustom ement-room-url-previews nil
+  "Show URL previews."
+  :type 'boolean)
+
+(defcustom ement-room-url-preview-prefix (concat "  "
+                                                 (char-to-string
+                                                  (char-from-name "BOX DRAWINGS HEAVY VERTICAL"))
+                                                 " ")
+  "String to insert before lines in URL preview."
+  :type 'string)
 
 (defcustom ement-room-header-line-format
   ;; TODO: Show in new screenshots.
@@ -2317,14 +2332,14 @@ the first and last nodes in the buffer, respectively."
 (defun ement-room--insert-event (event)
   "Insert EVENT into current buffer."
   (cl-labels ((format-event
-               (event) (format "TS:%S (%s)  Sender:%s  Message:%S"
-                               (/ (ement-event-origin-server-ts event) 1000)
-                               (format-time-string "%Y-%m-%d %H:%M:%S"
-                                                   (/ (ement-event-origin-server-ts event) 1000))
-                               (ement-user-id (ement-event-sender event))
-                               (when (alist-get 'body (ement-event-content event))
-                                 (substring-no-properties
-                                  (truncate-string-to-width (alist-get 'body (ement-event-content event)) 20))))))
+                (event) (format "TS:%S (%s)  Sender:%s  Message:%S"
+                                (/ (ement-event-origin-server-ts event) 1000)
+                                (format-time-string "%Y-%m-%d %H:%M:%S"
+                                                    (/ (ement-event-origin-server-ts event) 1000))
+                                (ement-user-id (ement-event-sender event))
+                                (when (alist-get 'body (ement-event-content event))
+                                  (substring-no-properties
+                                   (truncate-string-to-width (alist-get 'body (ement-event-content event)) 20))))))
     (ement-debug "INSERTING NEW EVENT: " (format-event event))
     (let* ((ewoc ement-ewoc)
            (event< (lambda (a b)
@@ -2333,12 +2348,13 @@ the first and last nodes in the buffer, respectively."
                         (ement-event-origin-server-ts b))))
            (node-before (ement-room--ewoc-node-before ewoc event event< :pred #'ement-event-p))
            new-node)
-      ;; HACK: Insert after any read markers.
+      ;; HACK: Insert after any read markers and link previews.
       (cl-loop for node-after-node-before = (ewoc-next ewoc node-before)
                while node-after-node-before
                while (pcase (ewoc-data node-after-node-before)
                        ;; MAYBE: Invert this and test that it's *not* an event node.
-                       ((or 'ement-room-read-receipt-marker 'ement-room-fully-read-marker) t))
+                       ((or 'ement-room-read-receipt-marker 'ement-room-fully-read-marker) t)
+                       ((app car-safe 'url-preview) t))
                do (setf node-before node-after-node-before))
       (setf new-node (if (not node-before)
                          (progn
@@ -2463,7 +2479,7 @@ If replaced event is not found, return nil, otherwise non-nil."
 To be used as the pretty-printer for `ewoc-create'.  THING may be
 an `ement-event' or `ement-user' struct, or a list like `(ts
 TIMESTAMP)', where TIMESTAMP is a Unix timestamp number of
-seconds."
+seconds. It is also used for printing url previews."
   ;; TODO: Use handlers to insert so e.g. membership events can be inserted silently.
   (pcase-exhaustive thing
     ((pred ement-event-p)
@@ -2483,7 +2499,9 @@ seconds."
     ((or 'ement-room-read-receipt-marker 'ement-room-fully-read-marker)
      (insert (propertize " "
                          'display '(space :width text :height (1))
-                         'face thing)))))
+                         'face thing)))
+    (`(url-preview ,(and (pred listp) url-preview))
+     (insert (ement-room--format-url-preview url-preview)))))
 
 ;; (defun ement-room--format-event (event)
 ;;   "Format `ement-event' EVENT."
@@ -2692,7 +2710,7 @@ If FORMATTED-P, return the formatted body content, when available."
                            (_ nil))))
     (when body
       ;; HACK: Once I got an error when body was nil, so let's avoid that.
-      (setf body (ement-room--linkify-urls body)))
+      (setf body (ement-room--linkify-urls body event)))
     ;; HACK: Ensure body isn't nil (e.g. redacted messages can have empty bodies).
     (unless body
       (setf body "[message has no body content]"))
@@ -2757,28 +2775,128 @@ ROOM defaults to the value of `ement-room'."
     ;; HACK: So we use the username slot, which was created just for this, for now.
     (when body
       (cl-macrolet ((matches-body-p
-                     (form) `(when-let ((string ,form))
-                               (string-match-p (regexp-quote string) body))))
+                      (form) `(when-let ((string ,form))
+                                (string-match-p (regexp-quote string) body))))
         (or (matches-body-p (ement-user-username user))
             (matches-body-p (ement-room--user-display-name user room))
             (matches-body-p (ement-user-id user)))))))
 
-(defun ement-room--linkify-urls (string)
-  "Return STRING with URLs in it made clickable."
+;; ;; ;; URLs
+
+(defun ement-room--linkify-urls (string msg)
+  "Return STRING with URLs in it made clickable.
+MSG is the message STRING was contained in."
   ;; Is there an existing Emacs function to do this?  I couldn't find one.
   ;; Yes, maybe: `goto-address-mode'.  TODO: Try goto-address-mode.
-  (with-temp-buffer
-    (insert string)
-    (goto-char (point-min))
-    (cl-loop while (re-search-forward (rx bow "http" (optional "s") "://" (1+ (not space)))
-                                      nil 'noerror)
-             do (make-text-button (match-beginning 0) (match-end 0)
-                                  'mouse-face 'highlight
-                                  'face 'link
-                                  'help-echo (match-string 0)
-                                  'action #'browse-url-at-mouse
-                                  'follow-link t))
-    (buffer-string)))
+  (let ((buffer (alist-get 'buffer (ement-room-local ement-room))))
+    (with-temp-buffer
+      (insert string)
+      (goto-char (point-min))
+      (cl-loop with c = 1
+               while (re-search-forward (rx bow "http" (optional "s") "://" (1+ (not space)))
+                                        nil 'noerror)
+               do (make-text-button (match-beginning 0) (match-end 0)
+                                    'mouse-face 'highlight
+                                    'face 'link
+                                    'help-echo (match-string 0)
+                                    'action #'browse-url-at-mouse
+                                    'follow-link t)
+               if (and ement-room-url-previews
+                       ;; spec recommends not getting previews for encrypted rooms
+                       (not (with-current-buffer buffer
+                              (cl-find-if (apply-partially #'equal "m.room.encryption")
+                                          (ement-room-invite-state ement-room)
+                                          :key #'ement-event-type))))
+               do (ement-room--get-preview buffer
+                                           msg
+                                           (buffer-substring-no-properties (match-beginning 0)
+                                                                           (match-end 0))
+                                           c)
+               (setf c (1+ c)))
+      (buffer-string))))
+
+(defun ement-room--get-preview (buf msg link num)
+  "Get preview of LINK. BUF is the ement room buffer this link is from.
+MSG is the message this link is from. NUM is the index for this link."
+  (with-current-buffer buf
+    (ement-api ement-session "preview_url"
+      :endpoint-category "media" :params `((url ,link))
+      :then (lambda (data)
+              (ement-room--insert-preview buf msg data num))
+      :else (lambda (why)
+              (ement-debug "Error in fetching preview: " why)))))
+
+(defun ement-room--insert-preview (buf msg data num)
+  "Insert link preview for DATA into MSG, if it isn't already present.
+BUF is the room buffer this link is from. NUM is the index for this link."
+  (cond ((alist-get 'errcode data) (message "Ement: You are being rate limited."))
+        (data (with-current-buffer buf
+                (let* ((node (ement-room--ewoc-node-before ement-ewoc
+                                                           msg
+                                                           (lambda (a b)
+                                                             (string= (ement-event-id a)
+                                                                      (ement-event-id b)))
+                                                           :pred #'ement-event-p))
+                       (next? (cl-loop with w = (ewoc-next ement-ewoc node)
+                                       repeat num
+                                       if (or (not w)
+                                              (not (listp (ewoc-data w)))
+                                              (not (eq (car (ewoc-data w))
+                                                       'url-preview))
+                                              (not (equal (cadr (ewoc-data w))
+                                                          data)))
+                                       return t
+                                       do (setf w (ewoc-next ement-ewoc w))
+                                       finally return nil)))
+                  (when next?
+                    (with-silent-modifications
+                      (ewoc-enter-after ement-ewoc
+                                        node
+                                        `(url-preview ,data)))))))))
+
+(defun ement-room--format-url-preview (data)
+  "Format a pretty-printed preview for DATA."
+  ;; Reference: https://ogp.me/
+  (let ((string ""))
+    ;; some pages don't respect the protocol and have missing fields, so we should check each one.
+    (when-let ((title (alist-get 'og:title data))
+               (url (alist-get 'og:url data)))
+      (setf string (concat (propertize title
+                                       'mouse-face 'highlight
+                                       'face '(link (:inherit ement-room-preview))
+                                       'help-echo (or url "no link")
+                                       'action (lambda (d)
+                                                 (when (stringp d)
+                                                   (browse-url d)))
+                                       'follow-link (when url t)
+                                       'button t
+                                       'button-data url
+                                       'keymap button-map
+                                       'category t)
+                           "\n")))
+    (when-let ((type (alist-get 'og:type data))
+               (typestr (cond ((string-match-p "music.*" type) "Music")
+                              ((string-match-p "video.*" type) "Video")
+                              ((string-match-p "article.*" type) "Article")
+                              ((string-match-p "book.*" type) "Book")
+                              ((string-match-p "profile.*" type) "Profile"))))
+      (setf string (concat string
+                           (propertize typestr 'face 'ement-room-preview)
+                           "\n")))
+    (when-let ((desc (alist-get 'og:description data)))
+      (let* ((current-left-margin (or (car (window-margins)) 0))
+             (current-right-margin (or (cdr (window-margins)) 0))
+             (available-width (abs (- (- (window-width)
+                                         (+ current-left-margin current-right-margin))
+                                      (length ement-room-url-preview-prefix)))))
+        (setf string (concat string
+                             (propertize (truncate-string-to-width (replace-regexp-in-string "\n"
+                                                                                             " "
+                                                                                             desc)
+                                                                   available-width)
+                                         'face 'ement-room-preview)
+                             "\n"))))
+    (propertize string 'line-prefix ement-room-url-preview-prefix)))
 
 ;; NOTE: This function is not useful when displaynames are shown in the margin, because
 ;; margins are not mouse-interactive in Emacs, therefore the help-echo function is called


alphapapa:
Thanks.  I like the way you render the previews with the bar to the left, it looks nice.

A few notes:

As I mentioned in the chat room, before this could be accepted, I'd need you to do the FSF copyright assignment, because I plan to submit Ement to GNU ELPA.  If you're uninterested or unable to do that, this could still be useful as a proof-of-concept.

For the implementation: Generally, we should avoid adding new types in the EWOC nodes, because that would require changes in several places that check node types and easily lead to new bugs.  So rather than inserting a new node for each preview, it would be simpler to handle it like reactions, by updating the message formatter to append any previews to the end of a message event.  The preview-getting callback would add the preview data to the event's `local` slot and invalidate the node in the EWOC, which would cause the message to be reformatted and display the preview at the end.  That would also naturally keep the previews with the message content they are for.

alphapapa:
As I mentioned in the chat room, what's probably needed is for me to implement a queue system in `plz` so that requests like this could be queued.

Regarding the current code in this PR: it's not quite what I had in mind.  It could be simpler and integrate more naturally into the existing code.  But I can't ask you to put more work into this until you've done the FSF CA, otherwise your work could be in vain, and I'll end up having to do it myself anyway.  So I don't think I should give more specific feedback until that's done.

alphapapa:
FYI, here's my WIP of using `plz` request queues:

* https://github.com/alphapapa/plz.el/commit/5a19fdb7c3f81d2bbd3dc4cadbb182da1556383e
* https://github.com/alphapapa/ement.el/commit/28edacd457b2e5a35097ac1081622d98bbba947c#diff-cb7286789d74818a389575302ea6c50460e5f39f23e6121843bf86261e37dc8eR1338

ement-room.el
```diff
@@ -2687,18 +2681,13 @@ Format defaults to `ement-room-message-format-spec', which see."
             (insert-and-inherit
              (propertize " "
                          'display `((margin right-margin) ,string))))))
-      (when-let* ((data (map-elt (ement-event-local event) (intern (concat "url-previews-"
-                                                                           (ement-event-id event)))))
-                  (num (or 1 (map-elt data
-                                      (intern (concat "url-previews-" (ement-event-id event)))))))
+      (when-let* ((data (ement-event-local event))
+                  (num (map-elt data 'url-preview-n)))
```
alphapapa, line 2685: Why store the number of previews?  They should probably just go in a list, and then...

ement-room.el
```diff
@@ -2687,18 +2681,13 @@ Format defaults to `ement-room-message-format-spec', which see."
...
         (goto-char (point-max))
-        (unless (looking-at-p "^\n")
-          (insert "\n"))
-        (cl-loop repeat num
-                 do (when-let ((inner-data (map-elt data (intern (concat "preview-"
-                                                                         (number-to-string num))))))
-                      (insert (ement-room--format-url-preview inner-data))))
-        )
+        (cl-loop for i from 1 to num
```
alphapapa, line 2687: ...rather than iterating through them like this, you just `(dolist (preview previews) ...)`.

ement-room.el
```diff
@@ -2798,6 +2787,9 @@ ROOM defaults to the value of `ement-room'."
 
 ;; ;; ;; URLs
 
+(defvar-local ement-room--preview-stop nil
```
alphapapa, line 2790: I appreciate your caution, but I don't think this complication is necessary.  If preview requests fail due to rate limiting (which should be very rare), they can simply fail.

ement-room.el
```diff
@@ -2817,151 +2809,113 @@ MSG is the message STRING was contained in."
                                     'action #'browse-url-at-mouse
                                     'follow-link t)
                if (and ement-room-url-previews
+                       (not (map-elt (map-elt (ement-event-content msg) 'm.relates_to)
```
alphapapa, line 2812: See `map-nested-elt`, as well as the many `pcase-let` examples in the code.

ement-room.el
```diff
@@ -2817,151 +2809,113 @@ MSG is the message STRING was contained in."
...
                        (not (with-current-buffer buffer
                               (cl-find-if (apply-partially #'equal "m.room.encryption")
                                           (ement-room-invite-state ement-room)
                                           :key #'ement-event-type))))
-               do (ement-room--get-preview buffer
-                                           msg
-                                           (buffer-substring-no-properties (match-beginning 0)
-                                                                           (match-end 0))
-                                           c)
+               do (let ((string (buffer-substring-no-properties (match-beginning 0) (match-end 0))))
```
alphapapa, line 2819: I feel like doing this in `--linkify-urls` is an abuse of the function.  :)  This function is intended to do one thing: return the string argument with any URLs in it having been linkified.

ement-room.el
```diff
@@ -2817,151 +2809,113 @@ MSG is the message STRING was contained in."
...
-;;                                        do (setf w (ewoc-next ement-ewoc w))
-;;                                        finally return nil)))
-;;                   (when next?
-;;                     (with-silent-modifications
-;;                       (ewoc-enter-after ement-ewoc
-;;                                         node
-;;                                         `(url-preview ,data)))))))))
-
-(defun ement-room--insert-preview (buf msg related data num)
+(defun ement-room--get-preview (msg link num)
```
alphapapa, line 2825: Allow me to get a bit nit-picky: names are important.  :)  `get` is an anti-pattern, IMHO.  Instead, just name the function for what it returns or does.  Please see the way other functions are named in this file.

As well, please follow the conventions around argument names (i.e. they should be clear and usually unabbreviated) and docstrings (e.g. "Return foo for BAR.").

ement-room.el
```diff
@@ -2817,151 +2809,113 @@ MSG is the message STRING was contained in."
...
+                                 if (ement--events-equal-p match e)
+                                 return e))
+         (saved-buffer (current-buffer)))
+    (unless (or ement-room--preview-stop
+                (map-elt (ement-event-local related-event)
+                         'url-preview-n))
+      (ement-api ement-session "preview_url"
+        :endpoint-category "media" :params `((url ,link))
+        :then (lambda (data)
+                (when (buffer-live-p saved-buffer)
```
alphapapa, line 2839: Rather than saving the buffer here, use the room struct itself in the lambda.  That way, if the room's buffer changes before the callback is called, it won't give an error.

ement-room.el
```diff
@@ -2817,151 +2809,113 @@ MSG is the message STRING was contained in."
...
+                  (with-current-buffer saved-buffer
+                    (ement-room--insert-preview msg related-event num (ement-room--cleanse-url-preview data link)))))
+        :else (lambda (why) (ement-debug "Error in fetching preview: " why))))))
+
+(defun ement-room--cleanse-url-preview (data link)
```
alphapapa, line 2844: For a function this small that is only called in one place, I'd prefer it be inlined.

Also, please leave a brief comment that explains why it's necessary.

ement-room.el
```diff
@@ -2817,151 +2809,113 @@ MSG is the message STRING was contained in."
...
+  "Ensure a url preview DATA always has a title and associated LINK."
+  (unless (alist-get 'og:url data) (push (cons 'og:url link) data))
```
alphapapa, line 2846: I am nitpicky about formatting, especially when it comes to consistency.  If the `push` on the following form is on its own line, this one should be too.

ement-room.el
```diff
@@ -2817,151 +2809,113 @@ MSG is the message STRING was contained in."
...
+  (unless (alist-get 'og:title data)
+    (push (cons 'og:title (replace-regexp-in-string "\\<https?://\\|/.*$\\|/$" "" link)) data))
```
alphapapa, line 2848: Would it be better to use something like `[untitled page]` than to reuse the URL?

ement-room.el
```diff
@@ -2817,151 +2809,113 @@ MSG is the message STRING was contained in."
...
+  data)
+
+(defun ement-room--insert-preview (msg related num data)
```
alphapapa, line 2851: Again, it's probably not necessary to pass an index around.  This is Lisp, not C.  :)  As well, being async code, the data could have changed since the request was made.

ement-room.el
```diff
@@ -2817,151 +2809,113 @@ MSG is the message STRING was contained in."
...
-                                                            (intern (concat "url-previews-" id)))
-                                                   'url-preview-n))
-                                most-positive-fixnum))
-               (cl-pushnew `(url-preview-n ,num)
-                           (map-elt (ement-event-local related)
-                                    (intern (concat "url-previews-" id)))))
-             (ewoc-invalidate ement-ewoc related-node))))))
+NUM is the index for this link."
+  (cond ((alist-get 'errcode data) (progn
+                                     (message "Ement: You are being rate limited. Waiting 10s before trying again.")
```
alphapapa, line 2855: I appreciate your thoroughness here, but let's leave retrying out of the code for now, to keep it simple.  These requests should fail rarely, and when they do, they aren't important.  I'd rather implement retries in a more general way at a higher level, e.g. in `plz` or `ement-api`.

ement-room.el
```diff
@@ -2817,151 +2809,113 @@ MSG is the message STRING was contained in."
...
+                                                       (lambda ()
+                                                         (with-current-buffer saved-room
+                                                           (setq-local ement-room--preview-stop nil)))))))
+        (data (let ((related-node (ement-room--ewoc-node-before ement-ewoc
+                                                                msg
+                                                                (lambda (a b)
+                                                                  (string= (ement-event-id a)
+                                                                           (ement-event-id b)))
+                                                                :pred #'ement-event-p))
+                    (key (intern (concat "preview-" (number-to-string num)))))
```
alphapapa, line 2868: Interning shouldn't be necessary for this.  If you need to do an alist lookup by string, do, e.g. `(alist-get KEY LIST nil nil #'equal)`.

ement-room.el
```diff
@@ -2817,151 +2809,113 @@ MSG is the message STRING was contained in."
...
+                          'action (lambda (d)
+                                    (when (stringp d)
+                                      (browse-url d)))
+                          'follow-link (alist-get 'og:url data)
+                          'button t
+                          'button-data (alist-get 'og:url data)
+                          'keymap button-map
+                          'category t)
+              "\n"))
+    (let* ((type? (alist-get 'og:type data))
```
alphapapa, line 2894: I don't mind the Scheme-style `foo?` naming in general, but Ement is an Emacs package, and I'm trying to follow Emacs's conventions, so please use, e.g. `foo-p` instead.

ement-room.el
```diff
@@ -2817,151 +2809,113 @@ MSG is the message STRING was contained in."
...
+           (special? (alist-get 'og:site_name data))
+           (typestr nil))
```
alphapapa, line 2896: Again, nitpicking, but please don't abbreviate unless really necessary.  How about just `type`?

ement-room.el
```diff
@@ -2817,151 +2809,113 @@ MSG is the message STRING was contained in."
...
+                            ((string-match-p "video.*" type?) "Video")
+                            ((string-match-p "article.*" type?) "Article")
+                            ((string-match-p "book.*" type?) "Book")
+                            ((string-match-p "profile.*" type?) "Profile"))))
+      (when typestr
+        (insert (propertize typestr 'face 'ement-room-preview) "\n")))
+    (when-let* ((desc (alist-get 'og:description data))
+                (current-left-margin (or (car (window-margins)) 0))
+                (current-right-margin (or (cdr (window-margins)) 0))
+                (available-width (abs (- (- (window-width)
```
alphapapa, line 2910: I don't think this code regarding window width and margins should be necessary.  Emacs should handle that kind of thing automatically.  As well, margins may not be used, depending on the user's configuration.  So let's keep this code as simple as possible.

alphapapa (changes requested):
Thanks for your work on this.  I made some comments on the code.

ement-room.el
rayes0: Yeah, that was mainly just for testing. I forgot to remove it :p

I'll probably just use dolist and use a variable as a counter or something to insert the correct order.

ement-room.el
rayes0: Would you suggest creating a separate filter function instead (though that would mean parsing the string for urls twice)? Maybe renaming linkify-urls as a more general url filter handler is a better option? Because it currently seems to be used only once in one place in the existing code, so...

ement-room.el
rayes0: Sure thing :+1:

ement-room.el
rayes0: Yeah, that was what I originally tried, but it seems plz doesn't call the function in the :then key in the same buffer as it was originally called in, instead it calls it in a *plz-request-curl* buffer. I considered filing a bug for plz, but I assumed it was desired behaviour.

I expect most users of plz would also expect that the function should be called in the original buffer plz was run in, so IMHO that should be changed on plz side.

ement-room.el
rayes0: It's to have a consistent preview, some previews may be missing the simplest of basic data like a URL or a title, which results in a preview that looks weird (eg: A preview with only a description, or maybe a preview with nothing at all).

Basically to ensure that all previews at least have a title and url associated with it. I should probably just inline the operations as you suggested though, since it's so small.

ement-room.el
rayes0: Yeah, now thinking about it, something like `[untitled page]` might be better. Currently it strips the URL to have only the domain name. Which one would you prefer to implement?

ement-room.el
rayes0: Hmm, how else would you suggest I do it? Because there can be multiple previews in the message, and it wouldn't know which one to insert if I don't pass it a number.

ement-room.el
rayes0: I don't think `equal` works for string-symbol comparisons? Only string-string or symbol-symbol I think.

I could use strings instead of symbols for all the preview keys but that feels...weird and unsettling :D I could do it that way if you wish though.

ement-room.el
rayes0: It's mostly just to keep the description a reasonable length. Some sites with poorly implemented previews have paragraphs of text as their description, which would all get inserted unless it's cropped :)

It currently already handles the case where margins are not used (it just uses left and right as 0).

A probably simpler and maybe better way though is to define a customization option for max length of descriptions and just truncate all descriptions to that length, but the window-width way is nice in that it is adaptive to the current size of the ement buffer, so for example if you have the ement room fullscreen as the only window, it will show enough preview of the description to fill that space, and it will get smaller as you decrease the size of the window.

I totally understand if you don't want to complicate things though, so I'm fine with whatever one you prefer.

ement-room.el
rayes0: This is because the number of actual needed URL previews doesn't necessarily correlate with the number of elements in the list. Some previews may have not been finished fetching yet, and sometimes homeservers seem to even send back duplicate previews if the connection is slow.

It isn't only used in the loop below either, also in other places to quickly get the number of expected previews, and also as a quick cheap check to see whether previews are present or not.

ement-room.el
alphapapa: This function does one thing.  If we need a function to do something else, it should be separate.

ement-room.el
alphapapa: I don't think that would work, because there is no guarantee that that buffer would still exist when the callback is called.  If it's necessary to call the callback in a certain buffer, it's the responsibility of the user to use `with-current-buffer` in the callback.

ement-room.el
alphapapa: I think I'd prefer something like "untitled page", because that would be more consistent and understandable than reiterating part of the URL.

ement-room.el
alphapapa: I'd suggest using an alist keyed on the URL rather than indexing them with numbers.

Basically, I think the way it should work is this: when receiving a preview for an event, add its data to that event's `local` slot, then invalidate that event in the EWOC.  When the event is rendered in the EWOC, render any previews to strings and append them to the event's rendered string.

ement-room.el
alphapapa: Correct, objects of different type are not `equal`.

I don't think we should intern more symbols to the obarray here if we can avoid it.  We should be able to key on the URL, I think.

ement-room.el
rayes0: I implemented it with alist keyed using URL at first, but then switched to keying by numbers is to provide information about the order. The first link in the message should have the first preview, the second should have the second, etc. Keying by URL alone would mean that previews would be randomly inserted depending on the order they were inserted into the alist, which would just be determined by the order they arrived from the network, which isn't great.

alphapapa: I agree that the preview order should match the order of URLs in the message, but rather than use indexes, we can simply parse the URLs out of the message in-order when appending the previews.  This is something that isn't going to happen often at all, and it probably won't even take 1 ms for the CPU, so performance isn't a concern and we can keep the code as simple as possible.

ement-room.el
```diff
@@ -2682,12 +2688,19 @@ Format defaults to `ement-room-message-format-spec', which see."
...
                          'display `((margin right-margin) ,string))))))
       (when-let* ((data (ement-event-local event))
-                  (num (map-elt data 'url-preview-n)))
+                  (previews (map-elt data 'url-previews)))
         (goto-char (point-max))
-        (cl-loop for i from 1 to num
-                 do (when-let ((inner-data (map-elt data (intern (concat "preview-" (number-to-string i))))))
-                      (unless (looking-at-p "^\n") (insert "\n"))
-                      (insert (ement-room--format-url-preview inner-data)))))
+        (dolist (preview previews)
```
alphapapa, line 2693: This is looking better.  Now it probably should go into its own function, similar to `ement-room--format-m.image`.  And the code that calls it should probably go in `ement-room--format-message-body`, treating it as a kind of "appendix" to the message body.

ement-room.el
```diff
@@ -2787,135 +2800,118 @@ ROOM defaults to the value of `ement-room'."
 
 ;; ;; ;; URLs
 
-(defvar-local ement-room--preview-stop nil
-  "Whether to proceed to previews for current room or not. Used in case of rate limit.")
-
-(defun ement-room--linkify-urls (string msg)
+(defun ement-room--linkify-urls (string message)
```
alphapapa, line 2803: Again, this function should not be overloaded to do anything regarding previews.  It shouldn't be necessary to change this function in this PR.

ement-room.el
```diff
@@ -2787,135 +2800,118 @@ ROOM defaults to the value of `ement-room'."
...
-                  (setf (alist-get 'url-preview-n (ement-event-local related)) num))
+                (when (not (alist-get 'url-previews (ement-event-local related)))
+                  (cl-pushnew (cons 'url-previews nil) (ement-event-local related)))
+                (unless (alist-get key (map-elt (ement-event-local related) 'url-previews))
+                  (cl-pushnew (cons key data) (map-elt (ement-event-local related) 'url-previews)))
                 (ewoc-invalidate ement-ewoc related-node)))))
 
 (defun ement-room--format-url-preview (data)
-  "Format a pretty-printed preview for DATA."
+  "Return a pretty-printed preview for DATA."
```
alphapapa, line 2878: Please mention what `DATA` is on a second line in the docstring (i.e. in this case I think it's the JSON from the parsed HTTP response).

ement-room.el
```diff
@@ -216,12 +216,18 @@ this one automatically.")
...
+                                                 (let ((char (char-from-name "BOX DRAWINGS HEAVY VERTICAL")))
+                                                   (if (char-displayable-p char)
+                                                       (char-to-string char)
+                                                     "|"))
                                                  " ")
-  "String to insert before lines in URL previews."
+  "Prefix to insert before lines in URL previews."
   :type 'string)
 
+(defcustom ement-room-url-preview-width 50
```
alphapapa, line 227: This could probably be a much larger number by default, like 200-250 characters.

ement-room.el
```diff
@@ -2787,135 +2800,118 @@ ROOM defaults to the value of `ement-room'."
...
-                      (ement-room--get-preview msg string c)))
+               do (with-current-buffer buffer
+                    (ement-room--url-preview message url c))
                (setf c (1+ c)))
       (buffer-string))))
 
-(defun ement-room--get-preview (msg link num)
-  "Get preview of LINK. MSG is the message this link is from. NUM is the index for this link."
-  (let* ((id (ement-event-id msg))
+(defun ement-room--url-preview (message link num)
```
alphapapa, line 2835: Please follow the conventions regarding argument names in the project.  For example, here `message` is actually an `ement-event` struct, so the argument should be named `event`.

alphapapa (changes requested):

ement-room.el
alphapapa: Basically, yes.  There should probably be a new function named something like `ement-room--preview-urls` that will receive the event as an argument, and when it finds URLs to preview in the event's body, it should call a function for each URL which requests the preview from the server and uses a callback to add the preview data to the event's local slot and invalidate the event's node when the room has a buffer.

We could, instead, fetch the previews when the event is formatted for the first time, so as to not fetch previews for events that are never going to be displayed (i.e. in rooms whose buffers aren't going to be shown).  That's probably better, in fact, but the rest of the procedure remains mostly the same.

I don't think the extra parsing will be a performance problem, but if it is, we can address it then.  We should avoid premature optimization.  :)

rayes0:
Sorry for not working on this in a while, was mainly waiting for the FSF CA to get done. Finally got it complete the other day.

ement-room.el
rayes0: I've currently left it how it is, mainly due to the fact that there is already a function named `ement-room--format-url-preview`. I think the most reasonable way to go about it is integrating the dolist into `ement-room--format-url-preview` and calling it from `ement-room--format-message-body`, but just wanted to confirm with you before I do it.

rayes0: I'll also probably implement the message parsing in this combined function to insert the previews in the correct order rather than just iterating using dolist.

alphapapa:
> Sorry for not working on this in a while, was mainly waiting for the FSF CA to get done. Finally got it complete the other day.

No problem.  That's great news.  I don't have time for another full review, but I'll make a few comments...

ement-room.el
```diff
@@ -2573,25 +2573,25 @@ Formats according to `ement-room-message-format-spec', which see."
   ;; TODO: Like other events, pop to a buffer showing the raw reaction events when a key is pressed.
   (if-let ((reactions (map-elt (ement-event-local event) 'reactions)))
       (cl-labels ((format-reaction
-                   (ks) (pcase-let* ((`(,key . ,senders) ks)
```
alphapapa, line 2555: Please remove any whitespace-only hunks from the patch.

ement-room.el
```diff
@@ -2800,41 +2803,44 @@ ROOM defaults to the value of `ement-room'."
...
+                                  'face 'link
+                                  'help-echo (match-string 0)
+                                  'action #'browse-url-at-mouse
+                                  'follow-link t
+                                  'ement-url t))
+    (buffer-string)))
+
+(defun ement-room--preview-urls (event rendered)
+  "Parse RENDERED body of EVENT for URLs and fetch previews for them."
+  (if (and (not (alist-get 'url-previews (ement-event-local event)))
```
alphapapa, line 2826: I'm not sure what `rendered` is here.

ement-room.el
```diff
@@ -2800,41 +2803,44 @@ ROOM defaults to the value of `ement-room'."
...
+           rendered)
+      (let ((room-buffer (alist-get 'buffer (ement-room-local ement-room))))
```
alphapapa, line 2828: If you can, please use `pcase-let` to destructure when possible.  I'm trying to use it consistently (although sometimes using `alist-get` is still necessary).

ement-room.el
```diff
@@ -2800,41 +2803,44 @@ ROOM defaults to the value of `ement-room'."
...
+        (with-temp-buffer
+          (insert rendered)
+          (goto-char (point-min))
+          (cl-loop with match = nil
+                   while (setf match (or (text-property-search-forward 'ement-url)
```
alphapapa, line 2833: Why not just do a regexp search for URLs on the message's plain-text body?  It will probably be faster, because Emacs's regexp engine is very well optimized, and the code will be much simpler.

ement-room.el
```diff
@@ -2847,35 +2853,35 @@ MESSAGE is the message STRING was contained in."
...
                                                          (push (cons 'og:url link) data))
                                                        (unless (alist-get 'og:title data)
                                                          (push (cons 'og:title "[untitled page]") data))
                                                        data)))))
         :else (lambda (why) (ement-debug "Error in fetching preview: " why))))))
 
-(defun ement-room--insert-preview (message related num data)
-  "Insert link preview for DATA into MESSAGE from node RELATED if it isn't already present.
-NUM is the index for this link."
+(defun ement-room--insert-preview (message related key data)
```
alphapapa, line 2866: Please write in terms of events rather than messages, because events are the logical unit we work in here.

ement-room.el
```diff
@@ -2847,35 +2853,35 @@ MESSAGE is the message STRING was contained in."
...
+  "Insert link preview for HTTP response DATA into MESSAGE from node RELATED,
```
alphapapa, line 2867: Note that docstrings' first lines should be a single sentence.  You can use `M-x checkdoc` to help with these guidelines.

ement-room.el
```diff
@@ -2847,35 +2853,35 @@ MESSAGE is the message STRING was contained in."
...
+if it isn't already present. KEY will be the key used for inserting."
```
alphapapa, line 2868: I don't understand what "KEY will be the key used for inserting" means.

ement-room.el
```diff
@@ -2847,35 +2853,35 @@ MESSAGE is the message STRING was contained in."
...
   (cond ((alist-get 'errcode data) (ement-debug "Rate limit in url preview"))
         (data (let ((related-node (ement-room--ewoc-node-before ement-ewoc
```
alphapapa, line 2870: Please avoid one-symbol-per-line formatting style.  This function has `(declare (indent defun))` specified to avoid this.  :)

ement-room.el
```diff
@@ -2847,35 +2853,35 @@ MESSAGE is the message STRING was contained in."
...
                                                                 (lambda (a b)
                                                                   (string= (ement-event-id a)
                                                                            (ement-event-id b)))
                                                                 :pred #'ement-event-p))
-                    (key (intern (concat "preview-" (number-to-string num)))))
-                (when (not (alist-get 'url-previews (ement-event-local related)))
-                  (cl-pushnew (cons 'url-previews nil) (ement-event-local related)))
-                (unless (alist-get key (map-elt (ement-event-local related) 'url-previews))
-                  (cl-pushnew (cons key data) (map-elt (ement-event-local related) 'url-previews)))
+                    (event (ement-event-local related)))
```
alphapapa, line 2876: If you're binding the event's `local` alist slot, the binding should be named `local`, not `event`--`event` should be the event struct itself.

But since you need to push to that alist anyway, you probably shouldn't bind it.  I may be mistaken, but I don't think setting the bound variable will also set the slot value; you need to set the place form's value (at least, that's been my experience).

ement-room.el
```diff
@@ -2847,35 +2853,35 @@ MESSAGE is the message STRING was contained in."
...
+                (unless (alist-get 'url-previews event)
```
alphapapa, line 2877: Please be consistent: use either `map-elt` or `alist-get`.  :)  (If performance matters, using `alist-get` will be faster, because `map-elt` is generic.  But see also `map-nested-elt`, which can make for clearer code than nested `alist-get`s, and it's fast enough--although I don't think `setf` works with it, yet.)

ement-room.el
```diff
@@ -2847,35 +2853,35 @@ MESSAGE is the message STRING was contained in."
...
+                  (cl-pushnew (cons 'url-previews nil) event))
+                (unless (alist-get key (map-elt event 'url-previews))
+                  (cl-pushnew (cons key data) (map-elt event 'url-previews)))
                 (ewoc-invalidate ement-ewoc related-node)))))
 
 (defun ement-room--format-url-preview (data)
-  "Return a pretty-printed preview for DATA."
+  "Return a pretty-printed preview for DATA, an alist of parsed JSON from the homeserver."
```
alphapapa, line 2884: Nitpick: To keep the first line of the docstring short, the second clause (which is just supporting information) can go in a second sentence on a second line.  (Checkdoc will complain, otherwise (although getting rid of all Checkdoc warnings is sometimes impractical)).

alphapapa (changes requested):
Thanks.

PR #64, opened 2022-04-15
Add/Change: Adaptive server name in prompt for ement-create-room

rayes0:
Quick change to show the homeserver for the selected session instead of matrix.org when creating rooms.


alphapapa:
I appreciate this idea.  It might be worth doing.  But first:

1.  The copyright assignment still needs to be done.  (Technically you could contribute up to 15 lines of changes to the project without doing so, but I don't want to have to keep track of that sort of thing.)
2. The indentation in this patch is very inconsistent.  Please use `(setf indent-tabs-mode nil)` and be sure to indent code properly before submitting (I recommend using `aggressive-indent-mode` for Lisp code, as it completely solves this problem).

alphapapa:
Well, I do appreciate your fixing the whitespace, but you also changed the whitespace in the entire file rather than just the lines you changed.  (Having worked on Ement in a different Emacs config for a while, which I neglected to set `indent-tabs-mode` in, it still has some tabs in it--I'll need to fix that in a separate commit sometime.)  Please update the patch to only have the lines you've added.

rayes0:
Should be good now, didn't realize you didn't me to fix the whole file :)

alphapapa:
Thank you.  :)

ement.el
```diff
@@ -382,11 +382,15 @@ new one automatically if necessary."
...
-  (interactive (list (ement-complete-session)
-		     :name (read-string "New room name: ")
-		     :alias (read-string "New room alias (e.g. \"foo\" for \"#foo:matrix.org\"): ")
-		     :topic (read-string "New room topic: ")
-		     :visibility (completing-read "New room type: " '(private public))))
+  (interactive (let ((session (ement-complete-session)))
+                 (list session
+                       :name (read-string "New room name: ")
+                       :alias (read-string (concat "New room alias (\"foo\" for \"#foo:"
+                                                   (replace-regexp-in-string "\\<https?://\\|/$" ""
```
alphapapa, line 389: I think we can just use `ement-server-name` here, right?

alphapapa (changes requested):

ement.el
```diff
@@ -382,11 +382,15 @@ new one automatically if necessary."
...
+                                                                             (ement-server-uri-prefix (ement-session-server session)))
+                                                   "\"): "))
```
alphapapa, line 391: I think we can just use `ement-server-name` here, right?

alphapapa (changes requested):

Issue #39, opened 2021-10-06
Should ement-room-image-keymap be a child of image-map?


vizs:
image-map has keys defined for scaling and rotating images.  I think
it would be handy to have these commands so we can resize/rotate images
in-buffer.  Org-mode, for instance, puts this keymap in the inline
image overlay.

ement-room-image-scale-mouse breaks after scaling the image with +/- though since
(image-property image :max-height) evaluates to nil.

alphapapa:
Maybe so.  If you'd like to test that out and report on the results, feel free.  :)

vizs:
Closing in favour of #67.

phil-s:
I suggest a call to `numberp` in place of the somewhat mysterious "impossible dummy value -1" which had me scratching my head wondering where that was set before I realised it was *purely* there to avoid an error from the `=` numeric equality check.

```diff
   (pcase-let* ((image (get-text-property pos 'display))
+               (max-height (image-property image :max-height))
                (window-width (window-body-width nil t))
                (window-height (window-body-height nil t))
-               ;; Image scaling commands set :max-height and friends to nil so use the
-               ;; impossible dummy value -1.  See <https://github.com/alphapapa/ement.el/issues/39>.
-               (new-height (if (= window-height (or (image-property image :max-height) -1))
+               ;; Image scaling commands set :max-height and friends to nil.
+               ;; See <https://github.com/alphapapa/ement.el/issues/39>.
+               (new-height (if (and (numberp max-height) (= window-height max-height))
```

phil-s:
Or just use `eql`, if it's *not* possible that it might be comparing an int with a float?  (I wasn't sure.)

alphapapa:
Ok, I'll merge a PR if you'll send it.

phil-s:
Addressed in https://github.com/alphapapa/ement.el/pull/255

PR #67, opened 2022-04-22
(ement-room-image-keymap) Inherit from image-map

vizs:
Continuation of #39.  I've been using this patch for a while now and no issues so far but note that I don't have images enabled for anything other messages (no avatars, room images, etc.) so there may be other issues that I do not know of.


alphapapa:
This may be a good change, but it does more than change the keymap.  :)

Also, if it's necessary to leave a comment explaining something, please try to make it self-contained rather than referring to a GitHub issue (that should generally only be done if the explanation is long or requires conversational context).  Or if it's in-between those lengths, you could put it in the commit message.

vizs:
I decided to go in between since someone might be interested in the (slightly) detailed analysis.

alphapapa:
@vizs Do you still plan to work on this?  i.e. since it changes more than what is mentioned in the commit message, it should be split into separate commits and PRs.

vizs:
This is basically done AFAIR.  I'm running ement with these modifications and haven't had any issues so far.  I will look into splitting these into separate PRs/commits sometime next week.

alphapapa:
Thanks.

PR #69, opened 2022-05-01
Change: Only highlight actual message body when replying

rayes0:
Something that became a problem while working on #63. Adds an `ement-body` text property set for the message body only so that when replying, only the actual message body is highlighted.

Before:
out

After:
out1


alphapapa:
This might be a good idea, but let's work on this after you've gotten the FSF CA done and the URL previews are merged.

PR #74, opened 2022-05-21
Correcting pantalaimon connection command

hochata:
It seems a botched edit changed the example connection command. I just restored the original version.


alphapapa:
Thanks for noticing and fixing that!  I don't know how that happened.  Recently I had a strange situation in which my Emacs undo list seemed to get corrupted and caused some weird corruption in a buffer; maybe it was related to that.  And then I obviously overlooked that line in the larger commit diff.  :)

PR #78, opened 2022-06-25
Add: Upload file by dragging them to an Emacs frame

vizs:
This commit allows one to upload a file to the current room by dragging and dropping it on an Emacs frame.

### DIFF
--- a/ement-room.el
+++ b/ement-room.el
@@ -48,6 +48,7 @@
 (require 'shr)
 (require 'subr-x)
 (require 'mwheel)
+(require 'dnd)
 
 (require 'ement-api)
 (require 'ement-lib)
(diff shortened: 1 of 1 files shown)


alphapapa:
Thanks, this seems like a good idea.

One minor problem I may forsee is that this doesn't check to see which window/buffer the file is dropped onto, so if the active window were a different buffer than the user intended to upload the file to, would it still work as expected?  (I haven't used the `dnd` library since I added support for this to `matrix-client` years ago, so I don't remember this detail about how it works.)

vizs:
25 Jun 2022, 19:16 by ***@***.***:

>
>
>
>
> Thanks, this seems like a good idea.
>
>
> One minor problem I may forsee is that this doesn't check to see which window/buffer the file is dropped onto, so if the active window were a different buffer than the user intended to upload the file to, would it still work as expected?  (I haven't used the > dnd>  library since I added support for this to > matrix-client>  years ago, so I don't remember this detail about how it works.)
>
>
Not too sure about what you mean here but if the user doesn't drop the file on the room window, then they won't be to upload the file.  E.g., if they drop it on a dired window, it will copy the file there; else it might be visited, etc.

If the current window is not a room buffer but if you drop the file onto an inactive window with the room buffer, then it uploads the file just fine.

I think it might be quicker to try it for yourself.

vizs:
Thanks!

PR #92, opened 2022-09-08
Emacs 27 compatibility fix #90

### LINKED ISSUE #90: Enable Emacs 27 fall-back for browse-url-handlers usage
```
;; Package-Requires: ((emacs "27.1") ...)
```

The `browse-url-handlers` option was added in Emacs 28.1:

```
** Browse URL

*** Added support for custom URL handlers.
There is a new variable 'browse-url-default-handlers' and a user
option 'browse-url-handlers' being alists with '(REGEXP-OR-PREDICATE
. FUNCTION)' entries allowing to define different browsing FUNCTIONs
depending on the URL to be browsed.  The variable is for default
handlers provided by Emacs itself or external packages, the user
option is for the user (and allows for overriding the default
handlers).

Formerly, one could do the same by setting
'browse-url-browser-function' to such an alist.  This usage is still
supported but deprecated.
```

phil-s:
If `browse-url-handlers` is not defined then fall back to using `browse-url-browser-function`.

### DIFF
--- a/ement-room.el
+++ b/ement-room.el
@@ -1917,11 +1917,17 @@ and erases the buffer."
         imenu-create-index-function #'ement-room--imenu-create-index-function
         ;; TODO: Use EWOC header/footer for, e.g. typing messages.
         ement-ewoc (ewoc-create #'ement-room--pp-thing))
-  (when (boundp 'browse-url-handlers)
-    ;; NOTE: This variable was added in Emacs 28.1.  In earlier versions, these links
-    ;; won't work within Ement.
-    (setq-local browse-url-handlers (cons (cons ement-room-matrix.to-url-regexp #'ement-room-browse-url)
-                                          browse-url-handlers)))
+  ;; Set the URL handler.  Note that `browse-url-handlers' was added in 28.1;
+  ;; prior to that `browse-url-browser-function' served double-duty.
+  (let ((handler (cons ement-room-matrix.to-url-regexp #'ement-room-browse-url)))
+    (if (boundp 'browse-url-handlers)
+        (setq-local browse-url-handlers (cons handler browse-url-handlers))
+      (setq-local browse-url-browser-function
+                  (cons handler
+                        (if (consp browse-url-browser-function)
+                            browse-url-browser-function
+                          (and browse-url-browser-function
+                               (list (cons "." browse-url-browser-function))))))))
   (setq-local completion-at-point-functions
               '(ement-room--complete-members-at-point ement-room--complete-rooms-at-point))  
   (setq-local dnd-protocol-alist (append '(("^file:///" . ement-room-dnd-upload-file)


phil-s:
I have not tested this yet.

alphapapa:
> I have not tested this yet.

Thanks for being clear about that.  Would you be able to test it?  I shouldn't merge it without it being tested first, and I don't have an Emacs 27 build handy.

phil-s:
Yep, I'll do some testing and follow-up; I just decided to open the PR at the time on the off-chance that would be comments in the interim.

phil-s:
I've tested this now (and fixed a bug).  I've pushed the fixed version to my fork.

FYI... I initially struggled with finding any `matrix.to` links on our server.  I ended up doing a text property search on some large histories, and was able to find and test some that way (even if they were all referencing something about 1-2 lines prior).

Most of the "in reply to" links from our Matrix server have URLs like this:

    https://DOMAIN1/#/room/!HASH1:DOMAIN2/$HASH2_HASH3-HASH4?via=DOMAIN2&via=DOMAIN3&via=DOMAIN4

(For our local domains, and what I presume are some kind of hashes or UUIDs.)

Notably, as well as the different domain, we have a `room/` component which doesn't match the regexp.

(Given this, I don't know why there were *any* `https://matrix.to/` URLs to be found, but I guess that's irrelevant.)

`ement-room-matrix\.to-url-regexp` is:

    "https?://matrix\\.to/#/\\([!#][^/]+\\)\\(?:/\\(\\$[^/?]+\\)\\)?\\(?:\\?\\([^z-a]+\\)\\)?"

Changing `matrix\\.to/#/` to `DOMAIN1/#/room/` made the "in reply to" links I'm seeing do the right thing (ement would jump to that location).  I.e.:

    "https?://DOMAIN1/#/room/\\([!#][^/]+\\)\\(?:/\\(\\$[^/?]+\\)\\)?\\(?:\\?\\([^z-a]+\\)\\)?"

I have no idea whether `room/` is a common component of these URLs or not, but perhaps that should be a optional part of the standard regexp, and then regexp made into a defcustom so that users can more easily tweak the initial domain part?

phil-s:
Oh, we also have these URLs:

    https://DOMAIN1/#/user/@USER:DOMAIN2

Perhaps in `matrix.to` URLs there's never a `user/` or `room/` or anything else, and it just figures it out?

alphapapa:
@phil-s 

> I've tested this now (and fixed a bug). I've pushed the fixed version to my fork. I was testing in both 27.2 and 28.1, and I believe this is working correctly for both.

Thank you!  I rebased and merged.

Regarding the other URL-related issues, I wasn't aware that the domain in those URLs was expected to be anything other than `matrix.to`.  If you would, please open another issue to discuss those problems and how we might solve them.

PR #100, opened 2022-09-21
Fix: (ement-notify-dbus-p) dbus-ping with timeout

### LINKED ISSUE #98: dbus-ping takes ~25s to finish evaluating
dbus-ping call in ement-notify takes ~25s in my laptop to finish evaluating.  In the meanwhile, Emacs is locked out and I cannot do anything :(

(benchmark-run 1 (dbus-ping :session "org.freedesktop.Notifications")) returns (25.001071378 0 0.0).

tsdh:
Use a 500ms timeout instead of the standard 25 seconds so that loading
ement-notify.el doesn't stall for ages when something DBUS is wonky.

This fixes #98.

### DIFF
--- a/README.org
+++ b/README.org
@@ -302,6 +302,7 @@ Note that, while ~matrix-client~ remains usable, and probably will for some time
 *Fixes*
 + Command ~ement-describe-room~ for rooms without topics.
 + Improve insertion of old messages around existing timestamp headers.
++ Reduce D-Bus notification system check timeout to 2 seconds (from the default of 25).
 
 ** 0.2.1
 
--- a/ement-notify.el
+++ b/ement-notify.el
@@ -52,7 +52,10 @@
   (and (featurep 'dbusbind)
        (require 'dbus nil :no-error)
        (dbus-ignore-errors (dbus-get-unique-name :session))
-       (dbus-ping :session "org.freedesktop.Notifications"))
+       ;; By default, emacs waits up to 25 seconds for a PONG.  Realistically, if there's
+       ;; no pong after 2000ms, there's pretty sure no notification service connected or
+       ;; the system's setup has issues.
+       (dbus-ping :session "org.freedesktop.Notifications" 2000))
   "Whether D-Bus notifications are usable.")
 
 ;;;; Customization


alphapapa:
Thanks.  I imagine this would work well in almost all cases, but I can imagine a case in which, e.g. the system just booted, Emacs has just launched, the system is thrashing the disk, and everything is taking longer than expected, which could cause it to take more than 500ms to respond, which would incorrectly disable the notifications.

What if, instead, we default the value to nil, and then do the ping asynchronously, setting the value to t if it succeeds?  What do you think?

tsdh:
Sounds good.  What async mechanism do you have in mind?  Just `make-thread`?  Can we rely on emacs being recent enough and compiled with thread support?

Or rather shell out and use a the classical `start-process`?  I think, `dbus-send --session --dest=org.freedesktop.Notifications /org/freedesktop/Notifications org.freedesktop.Notifications.ping` is the shell command equivalent to `dbus-ping` but I'm not sure.  @vizs, can you try out how that shell command exits (zero or not) and if it also stalls for a long time?

vizs:
```
% dbus-send --session --dest=org.freedesktop.Notifications /org/freedesktop/Notifications org.freedesktop.Notifications.ping
% echo $?
0
```

It finishes instantaneously.  While at the same time, Emacs' dbus-ping returns nil and takes 25s.

vizs:
I also tried the function 'n' to send a notification via dbus and it finished with no delay.

```
% n() {
	gdbus call -e -d "org.freedesktop.Notifications" \
		-o /org/freedesktop/Notifications \
		-m org.freedesktop.Notifications.Notify "" \
		"$1" "" "$2" "$3" "[]" "{}" 800
}
% n 200 asdf
(uint32 5,)
```

alphapapa:
@tsdh I wouldn't suggest using a thread for this, just a process.

@vizs 

> It finishes instantaneously. While at the same time, Emacs' dbus-ping returns nil and takes 25s.

That seems fishy to me.  I wonder if there's a bug in your system or config, or if there's something to be changed in `dbus-ping`.  That is, I wonder if there's anything that should be done in Ement at all.

vizs:
22 Sept 2022, 21:04 by ***@***.***:

>
> @vizs <https://github.com/vizs>
>
>
>>
>> It finishes instantaneously. While at the same time, Emacs' dbus-ping returns nil and takes 25s.
>>
>>
>
> That seems fishy to me.  I wonder if there's a bug in your system or config, or if there's something to be changed in > dbus-ping> .  That is, I wonder if there's anything that should be done in Ement at all.
>
>
I believe the number of people who can make sense of dbus is exactly zero, and those who tried to warned me not to poke at it...  Given that, and the fact that notifications work just fine despite the failing dbus-ping call, I think adding a modest TIMEOUT parameter is the only solution we have.

alphapapa:
@vizs 

> I believe the number of people who can make sense of dbus is exactly zero

That is true.  :)

> and those who tried to warned me not to poke at it...  Given that, and the fact that notifications work just fine despite the failing dbus-ping call, I think adding a modest TIMEOUT parameter is the only solution we have.

Okay, what if we just set the timeout to, say, 2 seconds?  Can you live with that as a user?  Then if we find out that it's happening to more people, we could try a more sophisticated solution like the async idea.  I just feel like 500ms is a bit too short.

vizs:
Sigh, my mail got dropped again.  Copying it in here,

>Okay, what if we just set the timeout to, say, 2 seconds? Can you live with that as a user? Then if we find out that it's happening to more people, we could try a more sophisticated solution like the async idea. I just feel like 500ms is a bit too short.

Fine by me.  ;-)

tsdh:
I've updated the PR to use a 2-sec timeout.

alphapapa:
@tsdh I updated the commit message to match the code.  :)

Thanks to all!

vizs:
>
> Okay, what if we just set the timeout to, say, 2 seconds?  Can you live with that as a user?  Then if we find out that it's happening to more people, we could try a more sophisticated solution like the async idea.  I just feel like 500ms is a bit too short.
>
>
Fine by me.  ;-)

PR #106, opened 2022-10-20
Add: ement-notify-notify-function

seblemaguer:
This provides the user to user his own notification handler

The wording of the documentation is not perfect because my brain is bit mushy right now, so don't hesitate to ping me if you want it to be rephrased.

### DIFF
--- a/README.org
+++ b/README.org
@@ -292,6 +292,11 @@ Note that, while ~matrix-client~ remains usable, and probably will for some time
 *Changes*
 + Improve ~ement-describe-room~ command (formatting, bindings).
 
+** 0.5.2
+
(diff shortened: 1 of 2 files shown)


alphapapa:
Hi Sébastien,

Allowing the user to use custom notification functions seems like a good idea.  It would probably be best to add a hook and use `run-hook-with-args`, so the user could add additional functions.

seblemaguer:
Hello,

sounds good to me for the hook. I will have to get a better look at this so. I move this PR in draft mode until I have something more "definite". 

For the FSF CA for Emacs, no I was not aware of this. I will check for that as well.

PR #112, opened 2022-11-08
Change `(setf (point) ...)` to `(goto-char ...)`

polaris64:
Emacs version 29.0.50 has deprecated the use of forms such as `(setf (point) n)`. The warning emitted is: -

```
Warning (bytecomp): ‘point’ is an obsolete generalized variable; use ‘goto-char’ instead.
```

### DIFF
--- a/ement-room.el
+++ b/ement-room.el
@@ -2525,7 +2525,8 @@ Also, mark room's buffer as unmodified."
   (interactive)
   (if-let ((fully-read-pos (when ement-room-fully-read-marker
                              (ewoc-location ement-room-fully-read-marker))))
-      (setf (point) fully-read-pos (window-start) fully-read-pos)
+      (progn (goto-char fully-read-pos)
+             (setf (window-start) fully-read-pos))
     ;; Unlike the fully-read marker, there doesn't seem to be a
     ;; simple way to get the user's read-receipt marker.  So if
     ;; we haven't seen either marker in the retrieved events, we


alphapapa:
Frankly, I do not understand why that deprecation was made, and I do not like it.  It's a useful paradigm, and I don't see any reason to throw it away.

Anyway, Emacs 29 is yet to be released.  So while I appreciate your keeping up with things, I don't want to merge this yet.  I'd like to find out why it was done and try to get the decision reversed.

polaris64:
OK that's absolutely fine. The main reason for this PR was to make you aware of the potential issue with Emacs 29 (in case you weren't aware already) and to provide a fix if it ends up needing to be applied.

alphapapa:
FYI, I just posted a message to emacs-devel asking for these generalized variable setters to be un-obsoleted.  (Not that I expect they will do so, but at least we can talk about it, I hope.)

And BTW, it's generally better to rebase your PRs than to merge `master` back into them.  It makes for a much simpler commit history.

polaris64:
> FYI, I just posted a message to emacs-devel asking for these generalized variable setters to be un-obsoleted. (Not that I expect they will do so, but at least we can talk about it, I hope.)

Thanks for letting me know, yes I just read the conversation so far this morning. I agree with you that the `setf` form is quite nice to use (makes it feel a bit more like Common Lisp) so it'll be interesting to see what they decide. I also didn't realise that marking something as "obsolete" was the first step in a discussion about whether something should be removed, I thought that had already been discussed and decided upon already!

> And BTW, it's generally better to rebase your PRs than to merge master back into them. It makes for a much simpler commit history.

OK no problem, I'll rebase this PR onto the current master in a bit.

PR #128, opened 2023-02-27
Add basic support for m.video events

viiru-:
This is very similar to the m.file handler (which would mostly work for videos as well), and is also missing support for thumbnails (however the events I'm seeing don't contain those anyway).

In comparison to ement-room--format-m.file this handler uses content.body as the caption instead of content.filename (which m.video doesn't contain, while m.file does have body). Also the video resolution is shown (<width>x<height>).

### DIFF
--- a/README.org
+++ b/README.org
@@ -302,6 +302,7 @@ Note that, while ~matrix-client~ remains usable, and probably will for some time
 + Option ~ement-room-timestamp-header-align~ controls how timestamp headers are aligned in room buffers.
 + Option ~ement-room-view-hook~ runs functions when ~ement-room-view~ is called.  (By default, it refreshes the room list buffer.)
 + In the room list, middle-clicking a room which has a buffer closes its buffer.
++ Basic support for video events.  (Thanks to [[https://github.com/viiru-][Arto Jantunen]].)
 
 *Changes*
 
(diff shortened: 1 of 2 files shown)


alphapapa:
Thanks, this looks good.  Have you done the FSF copyright assignment?

viiru-:
> Thanks, this looks good. Have you done the FSF copyright assignment?

No, I have not. I doubt this copy-paste job is technically copyrightable, but I can start the process for that. I have no idea how long that might take, though.

alphapapa:
> > Thanks, this looks good. Have you done the FSF copyright assignment?
> 
> No, I have not. I doubt this copy-paste job is technically copyrightable, but I can start the process for that. I have no idea how long that might take, though.

Personally, I agree that it's not "creative" enough on its own to be copyrightable separately, but AIUI the FSF has chosen this arbitrary number of lines to avoid having to make case-by-case judgement calls with their attorneys, and I've agreed to follow their guidelines for having this package in GNU ELPA.

The process is relatively painless in most jurisdictions.  I did it via email, but some places may require paperwork sent by mail.  In any case, it usually just takes a few weeks, but it depends on how busy the FSF secretary is.

I suppose an alternative would be for you to cancel this PR and make a feature request, and then I could write the code myself.  OTOH that seems like a sneaky workaround, but OTOH it's, as you said, a matter of copy-and-pasting my own code--there's only so many ways to write this kind of straightforward code, which is one of the selling points of Lisp and macros.  But I'm not a lawyer, so...  *shrug*

viiru-:
@alphapapa My FSF paperwork was finally accepted today, so if the technical side of this is acceptable it should now be ready to merge.

alphapapa:
Thanks!

PR #135, opened 2023-03-17
Add: Make timestamp headers optional

FrostyX:
I lurk on some low-traffic channels and there seem to be more timestamp headers than the actual messages. I want to disable them.

### DIFF
--- a/ement-room.el
+++ b/ement-room.el
@@ -483,6 +483,12 @@ Note that margin sizes must be set manually with
   "Number of messages to retrieve when loading earlier messages."
   :type 'integer)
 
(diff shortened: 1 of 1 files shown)


alphapapa:
Thanks for bringing this up.

This patch only adds one line, but it touches a lot of them, and there may be a simpler way: see the variable `ement-room-timestamp-header-delta`.

alphapapa:
AFAIK that would be the preferred solution, yes.

FrostyX:
Thank you,
in that case, I am closing this PR.

PR #151, opened 2023-05-24
Make it possible to seamlessly encrypt ement.el.gpg files on session save

daviwil:
This change came out of a discussion with @alphapapa where we were trying to figure out how to encrypt the Ement session file with the least amount of work.  This change was necessary to avoid a GPG key prompt every time the `ement.el.gpg` file gets saved.

### DIFF
--- a/ement.el
+++ b/ement.el
@@ -800,6 +800,11 @@ Returns nil if unable to read `ement-sessions-file'."
                                  :transaction-id transaction-id))))
     (message "Ement: Writing sessions...")
     (with-temp-file ement-sessions-file
+      ;; Ensure that `epa-file-encrypt-to' carries the correct value by making it buffer-local.
+      ;; This hack is inspired by this section of auth-sources.el:
+      ;; https://github.com/emacs-mirror/emacs/blob/73582ed6d4f227216d4c56c3f5f98c56e48796ca/lisp/auth-source.el#L1501-L1506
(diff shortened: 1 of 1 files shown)


alphapapa:
Thanks.  I'm slightly reluctant to add this, but I probably will.  :)

daviwil:
No worries if you're hesitant to!  The alternative is to provide a hook at the same location where someone can configure the `ement.el` buffer before it gets saved, then I can inject whatever behavior I want.

ement.el
```diff
@@ -865,18 +788,23 @@ Returns nil if unable to read `ement-sessions-file'."
...
+                                             :username username)
+                                 :server (list :name server-name
+                                               :uri-prefix uri-prefix)
+                                 :token token
+                                 :transaction-id transaction-id))))
     (message "Ement: Writing sessions...")
     (with-temp-file ement-sessions-file
+      ;; Ensure that `epa-file-encrypt-to' carries the correct value by making it buffer-local.
+      ;; This hack is inspired by this section of auth-sources.el:
+      ;; https://github.com/emacs-mirror/emacs/blob/73582ed6d4f227216d4c56c3f5f98c56e48796ca/lisp/auth-source.el#L1501-L1506
```
Thaodan, line 805: Does it make sense to link to Github when we can link to Gnu directly?

PR #152, opened 2023-05-24
Fix: avoid spurious indent in room membership changes

Stebalien:
Propertize the spaces in the widget's format string to hide them. We could, alternatively, use a zero-width space... but that won't work on a TTY.

This is a bit of a hack but it looks much nicer.

### DIFF
--- a/README.org
+++ b/README.org
@@ -308,6 +308,7 @@ Note that, while ~matrix-client~ remains usable, and probably will for some time
 
 *Fixes*
 + Minor adjustment to timestamp headers.
++ Extra indentation of some membership events.  (Thanks to [[https://github.com/Stebalien][Steven Allen]].)
 + Customization group for faces.
 
 ** 0.9.3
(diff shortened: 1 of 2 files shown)


Stebalien:
screen

alphapapa:
Hi Steven,

Thanks, this had been bugging for me for a while, but I hadn't thought of that clever solution.  :)

By the way, I like that screenshot.  Would you mind if I use it in the readme?  I'm planning to update the screenshots.

Stebalien:
Of course! Here's a trimmed down config as well:

```el
(use-package ement
  :custom
  (ement-room-timestamp-header-align 'center)
  (ement-room-timestamp-format "[%H:%M]")
  (ement-room-username-display-property '(raise -2.0))
  (ement-room-event-separator-display-property '(space :height 1.2))
  (ement-room-message-format-spec (concat (propertize "›" 'face 'shadow) "%L%B%r%R%t"))
  :custom-face
  (ement-room-timestamp ((t :weight light :height 0.8)))
  (ement-room-wrap-prefix ((t :inherit child-frame-border)))
  (ement-room-user ((t :family "URW Gothic" :bold t :height 1.2)))
  (ement-room-self ((t :family "URW Gothic" :bold t :italic t :height 1.2))))
```

PR #155, opened 2023-05-24
fix: only autoload when scrolling past the top

Stebalien:
Previously, this would attempt to retrieve old events whenever scrolling did nothing. Even when we were at the bottom of the buffer.

(I can update the changelog, but I don't think this warrants it)


alphapapa:
Hi Steven,

Thanks for this, but I don't fully understand the buggy behavior this fixes.  Could you explain in a bit more detail?

Stebalien:
The previous logic would load older events _either_ when scrolling past the bottom or scrolling past the top. The new logic only tries to load older events when scrolling past the top.

Stebalien:
For some context, this is a drive-by fix while trying to implement https://github.com/alphapapa/ement.el/issues/156 and unify the scroll logic. In _that_ case, a pixel scroll event can leave the window start unchanged (e.g., by scrolling less than a full line).

I just noticed that it was _technically_ wrong in this case as well, and submitted a fix.

alphapapa:
Well, how about that.  I guess I never scroll that far down in a buffer so as to notice this problem.  Thanks for catching this bug and fixing it.

PR #158, opened 2023-05-25
Fix sender header insertion with respect to membership changes

### LINKED ISSUE #157: Sender headers may need updating after receiving new/old events
Sender headers aren't displayed when the previous event isn't a message (e.g., it's a membership change event).

---

At the beginning of the buffer:

bob

(this one gets fixed by loading more messages)

---

After a timestamp header + some membership changes:

member-change

(the reply after the time change was sent by you but isn't attributed)

---

Another case. Everything after the timestamp isn't.

screen-3

Stebalien:
Only insert a sender header if the previous message is from a different sender, and there is no sender header between the current message and the previous message.

While the previous version of this function explicitly skipped the read marker, this version searches for known events (messages and sender headers), skipping everything it doesn't understand. This should make it robust to adding new events, widgets, etc.

fixes #157

### DIFF
--- a/ement-room.el
+++ b/ement-room.el
@@ -3014,63 +3014,43 @@ the first and last nodes in the buffer, respectively."
   "Insert sender headers into EWOC.
 Inserts headers between START-NODE and END-NODE, which default to
 the first and last nodes in the buffer, respectively."
-  (cl-labels ((read-marker-p (data)
-                (member data '(ement-room-fully-read-marker
-                               ement-room-read-receipt-marker)))
-              (message-event-p (data)
-                (and (ement-event-p data)
-                     (equal "m.room.message" (ement-event-type data))))
-              (insert-sender-before (node)
-                (ewoc-enter-before ewoc node (ement-event-sender (ewoc-data node)))))
-    (let* ((event-node (if (ement-event-p (ewoc-data start-node))
-                           start-node
-                         (ement-room--ewoc-next-matching ewoc start-node
-                           #'ement-event-p)))
-           (prev-node (when event-node
-                        ;; Just in case...
-                        (ewoc-prev ewoc event-node))))
-      (while (and event-node
-                  ;; I don't like looking up the location of these nodes on every loop
-                  ;; iteration, but it seems like the only reliable way to determine
-                  ;; whether we've reached the end node.  However, when this function is
-                  ;; called for short batches of events (or even a single event, like when
-                  ;; called from `ement-room--insert-event'), the overhead should be
-                  ;; minimal.
-                  (<= (ewoc-location event-node) (ewoc-location end-node)))
-        (when (message-event-p (ewoc-data event-node))
-          (if (not prev-node)
-              ;; No previous node and event is a message: insert header.
-              (insert-sender-before event-node)
-            ;; Previous node exists.
-            (when (read-marker-p (ewoc-data prev-node))
-              ;; Previous node is a read marker: we want to act as if they don't exist, so
-              ;; we set `prev-node' to the non-marker node before it.
-              (setf prev-node (ement-room--ewoc-next-matching ewoc prev-node
-                                (lambda (data)
-                                  (not (read-marker-p data)))
-                                #'ewoc-prev)))
-            (when prev-node
-              ;; A previous node still exists: maybe we need to add a header.
-              (cl-typecase (ewoc-data prev-node)
-                (ement-event
-                 ;; Previous node is an event.
-                 (when (and (message-event-p (ewoc-data prev-node))
-                            (not (equal (ement-event-sender (ewoc-data prev-node))
-                                        (ement-event-sender (ewoc-data event-node)))))
-                   ;; Previous node is a message event with a different sender: insert
-                   ;; header.
-                   (insert-sender-before event-node)))
-                ((or ement-user ement-room-membership-events)
-                 ;; Previous node is a user or coalesced membership events: do not insert
-                 ;; header.
-                 nil)
-                (t
-                 ;; Previous node is not an event and not a read marker: insert header.
-                 (insert-sender-before event-node))))))
+  (cl-labels ((message-event-p
+                (data) (and (ement-event-p data)
+                            (equal "m.room.message" (ement-event-type data)))))
+    ;; If we're not starting at a message node, forward to the previous message.
+    ;; If we don't find any messages, there's nothing to do anyways.
+    (when (and start-node (not (message-event-p (ewoc-data start-node))))
+      (setf start-node (ement-room--ewoc-next-matching ewoc start-node #'message-event-p)))
+
+    ;; Set the end node to the next message after the end node, if any. This:
+    ;; 1. Makes the loop check simple (we continue going until we run out of nodes or our
+    ;;    we find the end node).
+    ;; 2. Ensures we fix up sender headers after any inserted messages.
+    (when end-node
+      (setf end-node (ement-room--ewoc-next-matching ewoc end-node #'message-event-p)))
+
+    (let* ((event-node start-node) prev-node)
+      (while (and event-node (not (eq event-node end-node)))
+        ;; Scan back for the previous message or user header, if any.
+        (setf prev-node (ement-room--ewoc-next-matching ewoc event-node
+                          (lambda (data)
+                            (or (ement-user-p data) (message-event-p data)))
+                          #'ewoc-prev))
+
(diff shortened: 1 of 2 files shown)


alphapapa:
Thanks, I'll want to review and test this carefully before merging, so it might be a little while.  I do appreciate your working on this, since I don't use the "Elemental" style very often.

Stebalien:
np. Take your time.

Stebalien:
Hm. Actually, I've found one bug. The first sender header in the buffer _stays_ the first sender header, even after loading new events.

I believe `ement-room--insert-sender-headers` needs to widen its search by:

1. Scanning back from  `start-node` to the previous message.
2. Scanning forward from `end-node` to the next message.

Then it needs to "fix" all headers in this range. As you suggested, simply deleting the sender headers first then re-inserting them is likely the best approach.

Stebalien:
I'll tackle this tomorrow.

alphapapa:
> I believe `ement-room--insert-sender-headers` needs to widen its search by:
> 
>     1. Scanning back from  `start-node` to the previous message.
> 
>     2. Scanning forward from `end-node` to the next message.

IIRC it already does that--or maybe I did that for the timestamp headers.  I don't touch this code often.

> Then it needs to "fix" all headers in this range. As you suggested, simply deleting the sender headers first then re-inserting them is likely the best approach.

My hesitation with that approach is that it will, of course, take longer to do so with every incoming message.  Sometimes, e.g. I look way back in a room's history, loading a few thousand old events, which Ement handles just fine; but if this approach were used, it would likely perform poorly in that case.  So I'd prefer to avoid that if possible.

Something I haven't done is dig into the Matrix React SDK to see how it does it; but since it uses React, I don't know if any of its algorithms would be relevant here.

Maybe something relatively simple would help us ensure that we cover a large enough range while avoiding scanning the whole buffer, like scanning back to the previous full-date timestamp, or something like that.  Even that could be a performance issue in very busy rooms like #emacs:libera.chat or #matrix:matrix.org, but since the "Elemental" style isn't the default, maybe it would be acceptable.

Stebalien:
> My hesitation with that approach is that it will, of course, take longer to do so with every incoming message.

It shouldn't. I'm talking about fixing _just_ the range of messages that were modified, not the entire buffer.

> Maybe something relatively simple would help us ensure that we cover a large enough range while avoiding scanning the whole buffer, like scanning back to the previous full-date timestamp, or something like that

Ah, that's why I suggested scanning back/forward to the previous/next message (which we don't currently do for sender headers).

ement-room.el
```diff
@@ -3034,7 +3020,7 @@ Search starts from node START and moves by NEXT."
       ;; HACK: Insert after any read markers.
       (cl-loop for node-after-node-before = (ewoc-next ewoc event-node-before)
                while node-after-node-before
-               while (not (ement-event-p (ewoc-data node-after-node-before)))
+               while (read-marker-p (ewoc-data node-after-node-before))
```
Stebalien, line 3023: This change is optional, but it helps keep sender headers attached to their messages.

ement-room.el
```diff
@@ -2910,63 +2910,46 @@ the first and last nodes in the buffer, respectively."
...
+
+    ;; Set the end node to the next message after the end node, if any. This:
+    ;; 1. Makes the loop check simple (we continue going until we run out of nodes or our
+    ;;    we find the end node).
+    ;; 2. Ensures we fix up sender headers after any inserted messages.
+    (when end-node
+      (setf end-node (ement-room--ewoc-next-matching ewoc end-node #'message-event-p)))
+
+    (let* ((event-node start-node) prev-node)
+      (while (and event-node (not (eq event-node end-node)))
```
Stebalien, line 2932: `eq` should be sufficient now, as far as I can tell.

ement-room.el
```diff
@@ -2910,63 +2910,46 @@ the first and last nodes in the buffer, respectively."
...
+                          #'ewoc-prev))
+
+        (let ((sender (ement-event-sender (ewoc-data event-node))))
+          (cond
+           ;; No previous message/sender, insert a sender.
+           ((not prev-node) (ewoc-enter-before ewoc event-node sender))
+           ;; We have a previous sender, but it's wrong. Fix it.
+           ((ement-user-p (ewoc-data prev-node))
+            (unless (equal sender (ewoc-data prev-node))
+              (ewoc-set-data prev-node sender)))
```
Stebalien, line 2946: This can happen if we insert an event after a header but before another message. The read-marker fix below _should_ prevent this from happening (in most cases), but we might as well handle it.

ement-room.el
Stebalien: (I've tested this with simple printf debugging and it appears to work)

Stebalien:
Ok, I think this should actually be good now.

alphapapa:
Deferring this to 0.12 so 0.11 can be released sooner.

alphapapa:
@Stebalien Thanks for your patience with this.  I've kept deferring it because newer, more urgent bugs kept coming up that needed fixes released sooner.  I hope things are settling down now so we can start merging patches like this again.

alphapapa:
Thanks.  I was hoping to have time for things like this last weekend, but I didn't.  Probably I just need to start running this PR's code on my own system to test it "in the background."  FWIW, have you been using it regularly yourself?

Stebalien:
Not every day, but I use this patch whenever I use ement.el and haven't seen any issues.

Stebalien:
🥺

alphapapa:
> 🥺

Geez, you make me DDG that emoji to see what it's supposed to mean.  :P

You'll have to forgive me, as I have had a lot going on outside of my various Emacs packages.  And each one of them has a list of issues and PRs that are waiting.  In the past couple of days I just merged a couple that were 3 years old.

In the meantime, I would still be grateful for more feedback from having tested this patch.  

But I'll go ahead and rebase it and install the PR branch to test it myself...

alphapapa:
I've installed the PR's branch.  Seems to work well so far, but I'll have to remember to use the "Elemental" message format in order to see this code in action.

If no problems are found, I'll plan to merge this for v0.14.

Stebalien:
Thank you :heart:. And I understand, I've merged some _very_ old PRs.

In terms of testing, I've been using this more frequently and still haven't run into any issues.

alphapapa:
@Stebalien I think I'm ready to merge this.  

AFAIK you already signed the FSF CA, because I see some non-exempt commits from you in emacs.git, and of course you've contributed several here already.  But I can't find where I've asked you for confirmation already, so allow me to ask for confirmation now, just to be sure.  :)  Have you signed it already?

Thanks.

Stebalien:
Yep, the FSF officially owns all my Emacs work.

alphapapa:
> Yep, the FSF officially owns all my Emacs work.

Thanks.

BTW, I encountered an error which I've pushed an attempted fix for here, as well as adding a small failsafe: https://github.com/alphapapa/ement.el/pull/158/commits/b520bc6f95ce3559c02894267556ef67e7b105f2  Please let me know what you think about that change.

alphapapa:
Note as well that I discovered that the failsafe is unnecessary (which I should have realized from reading your comments in the code), so I removed it.  Seems to be working correctly now, but I still think it needs more testing, because this PR's code isn't the first time I've found a bug in this kind of code that only manifested after a few days (e.g. re-syncing after a long interruption).

Stebalien:
Hm. That shouldn't be possible. `prev-node` can be either `nil`, a user, or a message given the prior `(setf prev-node ...`.

alphapapa:
> Hm. That shouldn't be possible. `prev-node` can be either `nil`, a user, or a message given the prior `(setf prev-node ...`.

You're right, the bug was caused by my "Tidy..." commit that incorrectly consolidated one of the clauses.  I'll take care of it.  Thanks.

alphapapa:
I'm going to go ahead and merge this to master now.  Hopefully enough users use the "Elemental" format for it to get some testing.  :)

alphapapa:
@Stebalien Thanks for your work on this and your patience.

Stebalien:
No, thank you! I realize you don't use this feature so I'm just glad I don't have to maintain it myself.

alphapapa:
> No, thank you! I realize you don't use this feature so I'm just glad I don't have to maintain it myself.

I mean...if this breaks, I'll probably be pinging you to help fix it.  ;)

Stebalien:
No problem, that's better than constantly rebasing.

PR #159, opened 2023-05-26
Set the correct margins for visual-fill-column

Stebalien:
Otherwise, the margin text gets hidden when the window is too small and `visual-fill-column-mode` is enabled.

Given that `visual-fill-column` isn't a builtin package, feel free to ignore this patch. But it does make for _much_ more readable rooms by keeping the message width down, and this change shouldn't have any effect if the package isn't installed.

### DIFF
--- a/ement-room.el
+++ b/ement-room.el
@@ -217,6 +217,10 @@ In that case, sender names are aligned to the margin edge.")
 ;; Defined in Emacs 28.1: silence byte-compilation warning in earlier versions.
 (defvar browse-url-handlers)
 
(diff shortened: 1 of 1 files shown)


alphapapa:
I don't generally add code for third-party packages, but I'd be willing to add necessary hooks to enable this to be done in user config code.

Stebalien:
That won't require any changes to ement.el. The user can just use the room mode hook.

alphapapa:
Ok, good.  I'd be glad to add this to the documentation, then.

PR #162, opened 2023-05-27
Document message centering with visual-fill-column-mode

Stebalien:
I'm also happy to add a screenshot.

related to #159

### DIFF
--- a/README.org
+++ b/README.org
@@ -267,6 +267,24 @@ Emacs may not display certain symbols and emojis well by default.  Based on [[ht
   (set-fontset-font t 'unicode "Noto Emoji" nil 'append)
 #+END_SRC
 
+*** Wrapping message display
+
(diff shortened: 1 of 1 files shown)


alphapapa:
I would like to see a screenshot first, yes.  Thanks.

Stebalien:
Done.

(I also won't be offended if you feel that this isn't significant enough to include in the readme)

alphapapa:
Thanks, the screenshot helps.

> (I also won't be offended if you feel that this isn't significant enough to include in the readme)

I appreciate your saying that.  I do think that this kind of information is useful and should be somewhere, but maybe not in the readme/manual.  Maybe it would belong best in a wiki on the repo, or in a secondary documentation file showing how to integrate with other libraries.

If you don't mind, I'll sit on this for a while until we decide where best to put it.  Thanks.

Stebalien:
Yeah, that was my thought as well. The readme could quickly get cluttered that way.

Issue #175, opened 2023-06-27
Maintain position in room list buffer


phil-s:
Something I find frustrating is when I switch from the room list buffer to a room buffer, and then `q`uit back to the room list, and find that point has been moved back to the start of the room list.  This resetting seems to happen a lot.

I think the buffer is actually being erased and regenerated, so it's presumably not quite as trivial as a save-excursion; but to me it's an undesirable UX to have the apparent position forcibly changed this way.  Even if the buffer is being rebuilt, I think it would be beneficial to try to match the previous position based on context.  As it's a room list, "the line of the same room" would seem pretty easy and reliable.

alphapapa:
There is already code to do this, see: https://github.com/alphapapa/ement.el/blob/909abd44427ba1d3b9e83eda422c520dbe55880b/ement-room-list.el#L684-L693  Sometimes it doesn't work.  I haven't been able to find what is moving point back to the beginning sometimes, despite a lot of Edebug stepping and so forth.

phil-s:
WRT `(elisp)Quitting Windows`...

If I start in the room list buffer (in my test `(point)` is 690), and select a room with RET (replacing the room list buffer in its window), and then evaluate `(window-parameter nil 'quit-restore)` in the room buffer I see:

`(other (#<buffer *Ement Room List*> 1 #<marker at 1 in *Ement Room List*> 23) #<window 99 on *Ement Room: #emacs*> #<buffer *Ement Room: #emacs*>)`

Which is: `(METHOD (PREV-BUFFER PREV-WINDOW-START PREV-WINDOW-POINT HEIGHT) OWINDOW THIS-BUFFER)`

So `#<marker at 1 in *Ement Room List*>` is `PREV-WINDOW-START` and when I quit the room buffer it takes me back to position 1.

As that's a marker, if the text containing the marker is deleted the marker will be moved to the position at the beginning of the deleted region.  Hence if the whole buffer contents are erased, the marker will be moved to position 1.  This ties in with my observation about `ement-room-list-auto-update`.

The call to `set-window-point` won't help with that -- you would need to set the position of the specific marker used in the `quit-restore` property.

I believe `display-buffer-record-window` is responsible, in which I see:

```el
      ;; WINDOW shows another buffer.
      (with-current-buffer (window-buffer window)
	(set-window-parameter
	 window 'quit-restore
	 (list 'other
	       ;; A quadruple of WINDOW's buffer, start, point and height.
[...]

PR #193, opened 2023-07-28
Fix: (ement-room-list) Prevent window position being set via markers

phil-s:
Any markers in the room list buffer will be set to position 1 when `erase-buffer` is called each time the buffer is rebuilt, which caused the window point to be forcibly moved to the start of the buffer in many situations.

See #175

### DIFF
--- a/ement-room-list.el
+++ b/ement-room-list.el
@@ -668,12 +668,30 @@ DISPLAY-BUFFER-ACTION is nil, the buffer is not displayed."
                   column-sizes (cdr format-cons)
                   header-line-format (taxy-magit-section-format-header
                                       column-sizes ement-room-list-column-formatters))
-            (when-let ((window (get-buffer-window (current-buffer))))
+            (when-let ((window (get-buffer-window (current-buffer) t)))
               (setf window-point (window-point window)
                     window-start (window-start window)))
             (when ement-room-list-visibility-cache
               (setf magit-section-visibility-cache ement-room-list-visibility-cache))
             (add-hook 'kill-buffer-hook #'ement-room-list--cache-visibility nil 'local)
+            ;; Change relevant `quit-restore' markers to plain positions so they won't
+            ;; be reset to to 1 (on account of erasing the text containing the marker).
+            ;; See (info "(elisp) Quitting Windows") and `display-buffer-record-window'
+            ;; for `quit-restore', and (info "(elisp) Overview of Markers") regarding
+            ;; deleting text around a marker’s position.
+            (let ((roomlist (current-buffer)))
+              (cl-flet ((process (win)
+                                 (when-let* ((qr (window-parameter win 'quit-restore))
+                                             ((eq (car qr) 'other))
+                                             (obuffer (nth 1 qr))
+                                             ((eq (nth 0 obuffer) roomlist))
+                                             ((markerp (nth 2 obuffer))))
+                                   (setf (nth 2 obuffer)
+                                         (marker-position (nth 2 obuffer))))))
+                (walk-windows #'process t t)))
+            ;; For the same reasons, ensure that `switch-to-buffer' will not use
+            ;; a marker from `window-prev-buffers' for setting the window point.
+            (setq-local switch-to-buffer-preserve-window-point nil)
             ;; Before this point, no changes have been made to the buffer's contents.
             (delete-all-overlays)
             (erase-buffer)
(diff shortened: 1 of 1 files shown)


phil-s:
Ah, not quite done yet...

set-window-point(1): (set-window-point@my-debug apply set-window-point set-window-buffer-start-and-point switch-to-prev-buffer replace-buffer-in-windows kill-buffer funcall-interactively call-interactively command-execute)

phil-s:
Sigh, this looks like an Emacs bug...

`switch-to-buffer` only uses a marker from `window-prev-buffers` when `switch-to-buffer-preserve-window-point` is non-nil, but both `switch-to-prev-buffer` and `switch-to-next-buffer` call `set-window-buffer-start-and-point` using data from `window-prev-buffers`, and neither of those cases are conditional on that variable.

alphapapa:
> Sigh, this looks like an Emacs bug...
> 
> `switch-to-buffer` only uses a marker from `window-prev-buffers` when `switch-to-buffer-preserve-window-point` is non-nil, but both `switch-to-prev-buffer` and `switch-to-next-buffer` call `set-window-buffer-start-and-point` using data from `window-prev-buffers`, and neither of those cases are conditional on that variable.
> 
> (I've logged a bug report and CC'd you, Adam.)

Thank you, Phil.  This problem was driving me crazy.  I'm relieved to know that it's an Emacs bug after all.

phil-s:
The upstream bug report is https://debbugs.gnu.org/cgi/bugreport.cgi?bug=64911

alphapapa:
Maybe I'm missing something, but AIUI, the problem is that the window-point stuff uses markers internally, which I didn't realize; I expected that those were using integer positions.  So could we work around the problem by recording the point markers' positions and resetting from those (rather than using the actual markers)?

phil-s:
> could we work around the problem by recording the point markers' positions and resetting from those

Yes indeed -- and in fact I've done exactly that in the most recent commit.

> It might also be good to save the magit-section identity before erasing the buffer and try to go back to the same section after updating the buffer, if it's still present, and then fall back to positions.

Yep, I'd noted your `section-ident` code and had also figured that would be the next thing to do.  Completely on the same page.

alphapapa:
Thanks.  A final question, then: Do we still need to walk the windows and do things with their markers, or can we just record the marker positions for the room list window and use that?

phil-s:
We do still need to capture and reposition the markers, because we're not in control of the code which uses them to set the window point (in certain scenarios).  So we need to ensure that the markers are positioned the way we want them.

phil-s:
Can you confirm whether the pre-existing `section-ident` memory/restore is actually working for you?

I was just failing to make the related change to my code, and when I test the following interactively I get `nil`:

* `M-: (setq si (magit-section-ident (magit-current-section)))`
* refresh room list
* `M-: (magit-get-section si)`

alphapapa:
This may be relevant:
```
magit-section-ident-value is a byte-compiled Lisp function in
‘magit-section.el’.

(magit-section-ident-value OBJECT)

Return OBJECT’s value, making it constant and unique if necessary.

This is used to correlate different incarnations of the same
section, see ‘magit-section-ident’ and ‘magit-get-section’.

Sections whose values that are not constant and/or unique should
implement a method that return a value that can be used for this
purpose.


This is a generic function.

Implementations:

((section taxy-magit-section-section)) in ‘taxy-magit-section.el’.

Undocumented

((section magit-unpushed-section)) in ‘magit-log.el’.

"..@{push}" cannot be used as the value because that is
ambiguous if ‘push.default’ does not allow a 1:1 mapping, and
many commands would fail because of that.  But here that does
not matter and we need an unique value so we use that string
in the pushremote case.

((section magit-unpulled-section)) in ‘magit-log.el’.

"..@{push}" cannot be used as the value because that is
ambiguous if ‘push.default’ does not allow a 1:1 mapping, and
many commands would fail because of that.  But here that does
not matter and we need an unique value so we use that string
in the pushremote case.

((object eieio-default-superclass)) in ‘magit-section.el’.

Simply return the object itself.  That likely isn’t
good enough, so you need to implement your own method.

((section magit-section)) in ‘magit-section.el’.

Return the value unless it is an object.

Different object incarnations representing the same value then to
not be equal, so call this generic function on the object itself
to determine a constant value.
```
e.g. `(magit-section-ident-value (magit-current-section))` evaluates to the `[room session]` vector that is the value of the section at point.  That could probably be used to iterate over sections to find the one that matches the one that was at point before refreshing the buffer.

alphapapa:
> e.g. `(magit-section-ident-value (magit-current-section))` evaluates to the `[room session]` vector that is the value of the section at point. That could probably be used to iterate over sections to find the one that matches the one that was at point before refreshing the buffer.

Regarding that, this commit I merged changes the section idents from pointing to the actual room and session structs to using just their ID strings: https://github.com/alphapapa/ement.el/commit/2764c104efd94bd309e2ad7b249d5cafc6a45df7

alphapapa:
@phil-s I'm tidying up the list of issues for the v0.15 release.  Do you think this should be merged for v0.15 or deferred for v0.16?  (I generally try to keep down the number of changes that might introduce unexpected breakage, including ones I don't understand well.)

phil-s:
Let's leave this for now.  By accident it's ended up *not* being one of the patches I've been running locally, so I've not tested it recently and I don't want to have to review it at the moment.

alphapapa:
Sure, I'll mark it for v0.16 then.  Thanks.

PR #198, opened 2023-08-10
Update plz dependency

rdparker:
New installations of ement from elpa.gnu.org fail since plz 0.7 was pushed to the archive and ement requires 0.6.


rdparker:
Never commit before coffee.  A simple package-refresh-contents fixed it.

alphapapa:
I know that feeling...  :)

PR #201, opened 2023-08-11
Configurable emoji reaction picker

### LINKED ISSUE #199: New emoji input features in Emacs 29
@phil-s posted this in the room, which seems to be a very clever solution:
```el
(defun my-ement-room-send-common-reaction (key position)
  "Send a reaction."
  (interactive
   (list (minibuffer-with-setup-hook #'emoji-insert
           (read-string "Reaction: "))
         (point)))
  (ement-room-send-reaction key position))
```
> The (r)ecent sub-menu seems to cater for the "commonly used" case nicely enough (although I've not yet checked whether/how that is persisted between sessions). 

> Using emoji-search there also works, provided that enable-recursive-minibuffers is enabled.

oantolin:
I think this is ready to merge (and addresses #199), but note that it offers you the Emacs 29 emoji commands as options even if you don't have them. Likewise it binds them to keys. The docstring do mention you need Emacs 29 to use them, but maybe you'd want them to not even appear.

### DIFF
--- a/README.org
+++ b/README.org
@@ -299,6 +299,9 @@ Ement.el doesn't support encrypted rooms natively, but it can be used transparen
 
 *Additions*
 
++ Configurable emoji picker for sending reactions.  ([[https://github.com/alphapapa/ement.el/issues/199][#199]], [[https://github.com/alphapapa/ement.el/pull/201][#201]].  Thanks to [[https://github.com/oantolin][Omar Antolín Camarena]].)
+  - Option ~ement-room-reaction-picker~ sets the default picker.  Within that, the user may press ~C-g~ to choose a different one with a key bound in ~ement-room-reaction-map~.
+
 + A variety of enhancements for using compose buffers.  ([[https://github.com/alphapapa/ement.el/issues/140][#140]].  Thanks to [[https://github.com/phil-s][Phil Sainty]].)
   - Option ~ement-room-compose-buffer-display-action~ declares how and where a new compose buffer window should be displayed.  (By default, in a new window below the associated room buffer.)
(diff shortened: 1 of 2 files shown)


alphapapa:
Hi Omar,

Thanks, this looks very good.  Yes, I would like to have a version check in the option choices to only add ones from Emacs 29 when appropriate; I think we should avoid offering invalid choices to the user whenever possible.

oantolin:
Sure, I'll add the checks later today.

oantolin:
Done, I've added the checks. I did it individually for each command because (1) I dislike checking for versions since that seems like an indirect way to say what I mean, (2) maybe someone is running some in between Emacs commit that has some emoji commands but not others.

oantolin:
I left the mention to Emacs 29 in the docstring unconditionally, think of it as an advertisement. 😛

alphapapa:
@oantolin Finally merged.  Thanks for your work and your patience!

alphapapa:
@oantolin: Do you think we could make it easier for the user to select one of the other emoji input commands after pressing "C-g"? As it is I have to "describe-keymap ement-room-reaction-map" in another window to know what keys to press.

oantolin:
> Do you think we could make it easier for the user to select one of the other emoji input commands after pressing "C-g"?

Sure, I'd be happy to add a message showing the key bindings.

alphapapa:
> > Do you think we could make it easier for the user to select one of the other emoji input commands after pressing "C-g"?
> 
> Sure, I'd be happy to add a message showing the key bindings.

Thanks, that would be appreciated.

Issue #209, opened 2023-08-29
[Optimization] re-use buffers for message rendering


progfolio:
I noticed both `ement-room--render-html` and `ement-room--format-message` do their work in a temp buffer. Re-using a single buffer for each should reduce GC pressure and be  more performant.

alphapapa:
Thanks.

PR #210, opened 2023-08-29
Change: use dedicated buffer for message rendering, formatting

progfolio:
(ement-room--format-message):
(ement-room--render-html):
Prefer single buffer over allocating temp-buffer for each call. This should be perform better and reduce garbage.
See: #209


alphapapa:
Thanks.

alphapapa:
Merged in https://github.com/alphapapa/ement.el/commit/52faf4a4e2be330ebfa557e342b73d90a33545cb.  Can't push to the PR branch, so GitHub doesn't realize it.

PR #211, opened 2023-08-31
Add: (ement-set-avatar) Command to set user avatar

vizs:
As discussed in the matrix room, here is the PR :)


alphapapa:
Thanks.

Issue #223, opened 2023-09-24
ement-room-image-show unconditionally sets fullscreen frame-parameter


progfolio:
### OS/platform

Arch linux

### Emacs version and provenance

GNU Emacs 30.0.50 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.24.38, cairo version 1.17.8) of 2023-09-20
From arch user repository.

### Emacs command

emacs

### Emacs frame type

GUI

### Actions taken

1. Customize `display-buffer-alist` to prevent ement images from displaying in a new frame:

```emacs-lisp
  (add-to-list 'display-buffer-alist
               (cons "^\\*Ement image: "
                     (cons 'display-buffer-below-selected
                           '((inhibit-same-window . t)
                             (reusable-frames . nil)))))
```

2. Invoke `ement-room-image-show` on an image in any  room.

### Results

The altered `display-buffer-alist` rule is obeyed, but the current frame is resized.

### Expected results

The frame should retain it's dimensions.

### Etc.

ement.el @ commit:  52faf4a

In `ement-room-image-show`:

```emacs-lisp
    (pop-to-buffer new-buffer '((display-buffer-pop-up-frame)))
    (set-frame-parameter nil 'fullscreen 'maximized)))
```

would it be possible to set the frame parameter as part of the `ACTION` argument to `pop-to-buffer`?

alphapapa:
Probably the ACTION argument to `pop-to-buffer` should be in a `defcustom`, and the frame parameters should be passed in it via `pop-up-frame-parameters` (cf. Elisp manual 29.13.3).

alphapapa:
@progfolio Tag, you're it.  ;)

progfolio:
> Probably the ACTION argument to `pop-to-buffer` should be in a `defcustom`

I don't think this would be necessary if the ACTION specifies the parameters.
That part could still be overridden in display-buffer-alist.

> the frame parameters should be passed in it via `pop-up-frame-parameters` (cf.
> Elisp manual 29.13.3).

Thanks for hunting down that section of the manual.
Let me know what you think of #225.

alphapapa:
Merged the PR as-is.  Let me know if it needs anything else.  Thanks.

PR #225, opened 2023-09-26
fix: (ement-room-image-show) pass frame parameters

progfolio:


progfolio:
Sorry..."forg.el" made it too easy to fat finger this.

progfolio:
Come to think of it, why not just skip the ACTION argument and let the user decide?:

```emacs-lisp
(defun ement-room-image-show (pos)
  "Show image at POS in a new buffer."
  (interactive "d")
  (pcase-let* ((ement-event (ewoc-data (ewoc-locate ement-ewoc pos)))
               ((cl-struct ement-event id) ement-event)
               (buffer-name (format "*Ement image: %s*" id)))
    (unless (get-buffer buffer-name)
      (let ((image (copy-sequence (get-text-property pos 'display))))
        (when (fboundp 'imagemagick-types)
          ;; Only do this when ImageMagick is supported.
          ;; FIXME: When requiring Emacs 27+, remove this (I guess?).
          (setf (image-property image :type) 'imagemagick))
        (setf (image-property image :scale) 1.0
              (image-property image :max-width) nil
              (image-property image :max-height) nil)
        (with-current-buffer (get-buffer-create buffer-name)
          (erase-buffer)
          (insert-image image)
          (image-mode))))
    (pop-to-buffer buffer-name)))
```

alphapapa:
I think the current frame parameters should be the default, because I think most users will want that (to be able to see as much of a large image as possible), and no one will want to have to edit `display-buffer-alist` to achieve it.

So I still think [this](https://github.com/alphapapa/ement.el/issues/223#issuecomment-1732821966).

progfolio:
Won't have time to rework the patch in foreseeable future. Feel free to modify or discard.

alphapapa:
Merged as-is.  It's an improvement anyway.  Thanks.

PR #231, opened 2023-10-13
Add basic support for m.audio events

viiru-:
Very similar to the m.video handler (which is very similar to the m.file handler).

Again uses the message body as the title, and displays both audio duration and size.

### DIFF
--- a/ement-room.el
+++ b/ement-room.el
@@ -3508,6 +3508,7 @@ If FORMATTED-P, return the formatted body content, when available."
                            ("m.image" (ement-room--format-m.image event))
                            ("m.file" (ement-room--format-m.file event))
                            ("m.video" (ement-room--format-m.video event))
+                           ("m.audio" (ement-room--format-m.audio event))
                            (_ (if (or local-redacted-by unsigned-redacted-by)
                                   nil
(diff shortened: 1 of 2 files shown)


ement-room.el
```diff
@@ -4315,6 +4316,31 @@ Then invalidate EVENT's node to show the image."
...
+  "Return \"m.audio\" EVENT formatted as a string."
+  (pcase-let* (((cl-struct ement-event
+                           (content (map body
+                                         ('info (map mimetype duration size))
+                                         ('url mxc-url))))
+                event)
+               (url (when mxc-url
+                      (ement--mxc-to-url mxc-url ement-session)))
+               (human-size (file-size-human-readable size))
+               (human-duration (format-seconds "%m:%s" (seconds-to-time (/ duration 1000))))
```
alphapapa, line 4329: Is using `seconds-to-time` necessary?  The docstring of `format-seconds` says that it takes a SECONDS value as its argument.

alphapapa (changes requested):
Thank you.  If you'll check that feedback item, I'll update the changelog when I merge.

alphapapa (approved):

PR #239, opened 2023-11-04
Fix: (ement-room--format-message) %W considers only the current line

### LINKED ISSUE #182: Should the wrap prefix %W take newlines into account?
I'd been using the "IRC-style with right margin, with wrap-prefix" option for `ement-room-message-format-spec` but (on account of a user who had taken to putting rather lengthy status info in their 'friendly' username) wanted to stop the wrap prefix from following the username.

I decided I liked the idea of mimicing the Element formatting I was used to, where I had the username on top, the timestamp underneath on the left, and the message wrapping at the edge of the timestamp.  I figured that would be this:

    (setq ement-room-message-format-spec "%S:\n%t %W%B%r")

However the `%W` takes the entirety of the preceding string into account, and so the subsequent lines of the message are wrapped as if they had both the username and the timestamp on the left.

I don't imagine that would ever be useful when there are newlines involved, so should this be changed so that `%W` only considers the line it's in rather than the entire string prefix?

I tried to be sneaky by putting the wrap prefix at the start followed by a carriage return and then the 'real' value, but that just got me a `^M` in the output :)

phil-s:
When the format contains newlines, only the line containing %W is used to establish the width of the wrap-prefix.  Previously the widths of any preceding lines were included as well.

Closes #182.

### DIFF
--- a/README.org
+++ b/README.org
@@ -348,6 +348,7 @@ Ement.el doesn't support encrypted rooms natively, but it can be used transparen
 *Fixes*
 + Toggling images to fill the window body no longer triggers unintended scrolling.  (Thanks to [[https://github.com/phil-s][Phil Sainty]].)
 + Recognition of mentions after a newline.  ([[https://github.com/alphapapa/ement.el/issues/267][#267]].  Thanks to [[https://github.com/phil-s][Phil Sainty]].)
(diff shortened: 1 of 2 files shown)


phil-s:
Tested against all the default formats to confirm it has no adverse effects.

phil-s:
AFAIK, this one is good to go.

phil-s:
Rebased over current master.

phil-s:
Rebased over current master.

alphapapa:
Thanks!

PR #232, opened 2023-10-14
Expand the E2EE documentation

telenieko:
Thanks to alphapapa & contributors for such an amazing Matrix client for Emacs. Combined with Beeper I now have all my social networks in one place: Emacs!

We now include as much information as possible about E2EE set up in a separate file (e2ee.org) in order to:

- Make up for the lack of documentation from Pantalaimon
- Have a place to include tips & tricks about making it work
- Have a place to direct users asking about E2EE in Ement.el (and suggesting them to submit updates/additions), like:
  - https://github.com/alphapapa/ement.el/issues/54
  - https://github.com/alphapapa/ement.el/issues/60
  - https://github.com/matrix-org/pantalaimon/issues/157
  - https://github.com/alphapapa/ement.el/issues/33

Fixes #153

PR #242, opened 2023-11-23
File to explain how to configure Pantalaimon

dionisos2:
Add the using-pantalaimon.md file, to explain how to configure Pantalaimon.

### DIFF
--- a/using-pantalaimon.md
+++ b/using-pantalaimon.md
@@ -2,10 +2,10 @@
 
 Ement.el doesn't support encrypted rooms natively, but it can be used transparently with the E2EE-aware reverse proxy daemon [Pantalaimon](https://github.com/matrix-org/pantalaimon/).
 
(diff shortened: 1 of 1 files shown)


alphapapa:
Hi @dionisos2,

Thanks, this looks like the most comprehensive guide I've seen so far.

Have you seen this other PR, https://github.com/alphapapa/ement.el/pull/232 ?  It would probably be best if we collaborated on a single document that consolidates all of these instructions.

dionisos2:
Hi @alphapapa, and thank you.

I saw this PR, but I just opened all the links, and then missed it was actually a PR, and so I missed the [e2ee.org](https://github.com/telenieko/ement.el/blob/doc-e2ee/e2ee.org) document.

I strongly agree that we should merge the information on both documents and collaborate on a single one.

Issue #240, opened 2023-11-07
Constrain the ement-room--event-mentions-user-p regexp


phil-s:
At present the word `uphill` in a message is treated as mentioning the username `phil` :)

Let's at minimum wrap the pattern with word boundaries?

alphapapa:
Makes sense.  Thanks.

phil-s:
I looked to see if there were restrictions on characters used for a user's display name, and I'm not seeing anything.  Searching for `displayname` at https://spec.matrix.org/v1.8/client-server-api/ gives lots of info, but it seems like the value may be "any string".

So we can't simply wrap the quoted display name with word boundaries, but I suspect that even if the user had non-word characters at the start or end of their display name, we *could* assume that any mention of their name will have non-word characters (or bol/eol) *surrounding* the display name.  So maybe something like this:

    "\\(?:^\\|[^[:word:]]\\)\\(DISPLAYNAME\\)\\(?:$\\|[^[:word:]]\\)"

Possibly in the form of a `rx`-based template in a new variable or user option.

alphapapa:
It's kind of hidden here: https://spec.matrix.org/v1.8/appendices/#user-identifiers

phil-s:
Ah, yes, we have different rules for username and displayname (I was only thinking about displayname).

At present `ement-room--event-mentions-user-p` uses the same code for both.  In fact it matches on three names... 

        (or (matches-body-p (ement-user-username user))
            (matches-body-p (ement--user-displayname-in room user))
            (matches-body-p (ement-user-id user)))))))

Where `username` is the "Username part of user's Matrix ID." in the user struct; so I guess we apply the character constraints from your spec URL to both `ement-user-username` and `ement-user-id`, and only `ement--user-displayname-in` should be the more permissive one.

Konubinix:
@phil-s how did you mitigate that issue ?

phil-s:
> Are there any downsides to Phil's example of adding word boundaries to the regexp?

It's not *generally* correct (per the discussion earlier in this issue), but it's going to be better than the current code for most users, so merging a PR for that sounds like a good idea until someone gets around to writing a proper fix.

Konubinix:
If that's ok, I will suggest a PR with the mitigation that @phil-s  and I are using, as it was suggested by @phil-s and approved by @alphapapa .

Konubinix:
As suggested, I started a pull request. I'd be happy to amend it to follow your recommendations @alphapapa

[...]

PR #244, opened 2023-11-27
fix: Make sure the matched string is not simply a substring of a bigger word

Konubinix:
If one displayname is 'sam' and another's nickname is 'samy', then all events to samy were recognized as mentionning sam as well.

I figured this out because my displayname is sam and all events with the word same where found as mentionning me.


phil-s:
See https://github.com/alphapapa/ement.el/issues/240

alphapapa:
Yes, please address this with #240.

PR #245, opened 2023-11-29
Support variable-pitch face for plain text messages

### LINKED ISSUE #174: Improving support for variable-pitch text
I'd like to be able to configure a variable-pitch font for all regular text in a chat room, while seeing a fixed-width font for any pre-formatted text.

At present I can see a user option `ement-room-shr-use-fonts` which results in formatted (org.matrix.custom.html) messages being variable-pitch, but (as the docstring points out), "some messages won’t display in the same font as others", and seeing fixed-width messages interspersed with the formatted ones ends up being fairly ugly and distracting.

It would be nice if there was an option for un-formatted text to be rendered with the same variable-pitch face.

I can customize the `ement-room-message-text` face, but experimentally this causes pre-formatted code blocks (sent via the org-mode support) to render variable-pitch (although in-line verbatim and code text does end up fixed-width).

Without that face change I think 'shr-use-fonts' was doing all the right things for formatted messages, so a way to target *only* "unformatted" messages could work, but there might be some cleaner approach to unifying the behaviour.

phil-s:
Closes #174.

### DIFF
--- a/ement-room.el
+++ b/ement-room.el
@@ -870,10 +870,10 @@ the original messages visible."
   :type 'boolean)
 
 (defcustom ement-room-shr-use-fonts nil
-  "Enable `shr' variable-pitch fonts for formatted bodies.
-If non-nil, `shr' may use variable-pitch fonts for formatted
-bodies (which include most replies), which means that some
-messages won't display in the same font as others."
+  "Enable `shr' variable-pitch fonts for message bodies.
+If non-nil, `shr' may selectively use variable-pitch fonts for
+formatted message bodies (which includes most replies), and plain
+text messages will also be displayed in a variable-pitch font."
   :type '(choice (const :tag "Disable variable-pitch fonts" nil)
                  (const :tag "Enable variable-pitch fonts" t)))
 
@@ -1309,7 +1309,7 @@ spec) without requiring all events to use the same margin width."
   ;; This used to be a macro in --format-message, which is probably better for
   ;; performance, but using a function is clearer, and avoids premature optimization.
   (pcase-let* (((cl-struct ement-event sender
-                           (content (map msgtype))
+                           (content (map msgtype format ('m.new_content new-content)))
                            (unsigned (map ('redacted_by unsigned-redacted-by)))
                            (local (map ('redacted-by local-redacted-by))))
                 event)
@@ -1336,7 +1336,16 @@ spec) without requiring all events to use the same margin width."
                                               (color-darken-name message-color ement-room-prism-message-lightening))))))))
                (redacted-face (when (or local-redacted-by unsigned-redacted-by)
                                 'ement-room-redacted))
-               (body-face (list :inherit (delq nil (list redacted-face context-face type-face)))))
+               ;; For visual consistency, apply the variable-pitch `shr-text' face to
+               ;; non-HTML messages when `ement-room-shr-use-fonts' is non-nil (HTML
+               ;; messages are fontified by shr itself).
+               (shr-text-face (when (and ement-room-shr-use-fonts
+                                         (not (equal (or format (alist-get 'format new-content))
+                                                     "org.matrix.custom.html")))
+                                ;; The `shr-text' face was added in Emacs 29.1.
+                                (or (and (facep 'shr-text) 'shr-text)
+                                    'variable-pitch)))
+               (body-face (list :inherit (delq nil (list redacted-face context-face type-face shr-text-face)))))
     (if prism-color
         (plist-put body-face :foreground prism-color)
       body-face)))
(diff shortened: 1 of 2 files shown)


phil-s:
Marking as a draft pending more testing time, in case I spot any additional special cases.

phil-s:
I'll need to deal with 'emote' messages too:

```plain
 (:content
  (body . "<message that was rendered in monospace>")
  (m.mentions)
  (msgtype . "m.emote"))
 (:origin-server-ts . 1701224341051)
 (:type . "m.room.message")
```

I'm not sure whether to extend the `(equal msgtype "m.text")` test, or if it's better to drop it entirely.

phil-s:
I've tested a bunch of different things and not spotted any more issues, so this PR is no longer a Draft.

phil-s:
Force-pushed to squash the prior fixup commit.

phil-s:
Bah, and *now* I spot something else to tidy up...

phil-s:
Have done more testing, and am happy with this now.

ement-room.el
```diff
@@ -988,7 +991,16 @@ spec) without requiring all events to use the same margin width."
                                               (color-darken-name message-color ement-room-prism-message-lightening))))))))
                (redacted-face (when (or local-redacted-by unsigned-redacted-by)
                                 'ement-room-redacted))
-               (body-face (list :inherit (delq nil (list redacted-face context-face type-face)))))
+               ;; For visual consistency, apply the variable-pitch `shr-text' face to
+               ;; non-HTML messages when `ement-room-use-variable-pitch' is non-nil.
+               ;; (HTML messages are fontified by shr itself.)
+               (shr-text-face (when (and ement-room-use-variable-pitch
+                                         (not (equal (or format (alist-get 'format new-content))
```
alphapapa, line 998: Nitpick: For consistency, would you bind the new-content's format in the pcase-let form?

phil-s: Could do.  I thought it made sense to defer it to here for efficiency, as it saves a bit of unnecessary processing in the cases where it's not being used, which is the vast majority of the cases.

phil-s:
> How do you think we should describe it in the changelog?

How about:

Option: `ement-room-use-variable-pitch` (previously named `ement-room-shr-use-fonts`) enables variable-pitch fonts for all message types.  (This option previously supported formatted messages, but now works for plain text messages as well.)

ement-room.el
alphapapa: Good point, I agree.  Thanks.

alphapapa (approved):

alphapapa:
Thanks.  If you don't mind updating the changelog, then I'll merge this.

phil-s:
As another micro optimisation, seeing as how the following won't ever vary in a given session:

```el
(or (and (facep 'shr-text) 'shr-text)
    'variable-pitch))
```

I thought I might `defconst ement-room-variable-pitch-face` with that value.

alphapapa:
> As another micro optimisation, seeing as how the following won't ever vary in a given session:
> 
> ```emacs-lisp
> (or (and (facep 'shr-text) 'shr-text)
>     'variable-pitch))
> ```
> 
> I thought I might `defconst ement-room-variable-pitch-face` with that value.

Yeah, I think that makes sense.  You might need to verify that `shr` is loaded first.

phil-s:
> You might need to verify that shr is loaded first.

That's all good -- ement-room.el has a `(require 'shr)`.

phil-s:
I'm investigating an issue which is probably connected to this in some way.  Actual rendering appears fine, but something in my config is currently triggering a lot of face-attribute-related messages; so don't merge this in the interim, just in case there's a bug to fix.

phil-s:
Yep, that last comment/issue was totally unrelated.  Nothing to see here!

I was seeing floods of this in Messages:

```plain
Invalid face attribute :background nil
Invalid face reference: ement-room-mention
Invalid face attribute :inherit (ement-room-mention ement-room-message-text)
```

And it took me a stupidly long time to recognise the literal-ness of the first message.

```el
;; Bad:
(face-remap-add-relative 'ement-room-mention :background nil)
;; Good:
(face-remap-add-relative 'ement-room-mention :background 'unspecified)
```

phil-s:
AFAIK, this one is good to go.

alphapapa:
I'd like to release v0.14 for tomorrow, so I'll merge this for v0.15 afterward, so it can be tested on master for a while.

phil-s:
Rebased over current master.

phil-s:
Rebased over current master.

alphapapa:
Merged.  Thanks for your work and your patience.

PR #246, opened 2023-12-02
Fix: (ement-room-send-reaction) Bail out early when there's no event at point

phil-s:
Don't read the reaction from the user if it can't be used.

### DIFF
--- a/README.org
+++ b/README.org
@@ -311,6 +311,7 @@ Ement.el doesn't support encrypted rooms natively, but it can be used transparen
 + Insertion of sender headers (when using "Elemental" message format).  (Refactoring contributed by [[https://github.com/Stebalien][Steven Allen]].)
 + Some room event data was being unintentionally serialized to disk when caching the room list visibility state. ([[https://github.com/alphapapa/ement.el/issues/256][#256]])
 + Notifications buffer restores properly when bookmarked.
++ Command ~ement-room-send-reaction~ checks for an event at point.  (Thanks to [[https://github.com/phil-s][Phil Sainty]].)
 
 ** 0.13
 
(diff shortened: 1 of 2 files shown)


alphapapa:
Hm, I think probably the function should be refactored to only take the key and event as arguments, and all the reading and temporary highlighting should be done in the interactive form (if possible--maybe there's a reason I didn't do it that way already).  What do you think?  Thanks.

phil-s:
Offhand it sounds sensible to me.  We'd just need to call it out as an API change?

`ement-room-toggle-reaction` has a call to `(ement-room-send-reaction key (point))` but I don't see a reason for that to not pass the event at point (which it looks like it already has as its `event` argument, and I don't see any non-interactive calls to that toggle function).  So I'm not seeing any immediate reason to avoid refactoring.

alphapapa:
> Offhand it sounds sensible to me. We'd just need to call it out as an API change?

Even though it's not a double-dashed function, I don't think we need to record the API change, because I don't think anyone else is calling these functions from their own code.  But if you want to, I don't object.


> `ement-room-toggle-reaction` has a call to `(ement-room-send-reaction key (point))` but I don't see a reason for that to not pass the event at point (which it looks like it already has as its `event` argument, and I don't see any non-interactive calls to that toggle function). So I'm not seeing any immediate reason to avoid refactoring.

Yeah, I think it will work.  I'm still not sure why I did it that way in the first place; maybe something to do with toggling using the buttons.

alphapapa:
In the long run, this command should probably be slightly refactored to take the room, session, and event as arguments, and provide them all interactively (so e.g. reactions could be sent from the notifications buffer).  But until then, this is still an improvement, so I'll merge it for v0.14.

PR #254, opened 2024-01-21
Tidy: Use custom docstring for ement-room-mode-hook

phil-s:
There are mode line constructs in the mode-name for `ement-room-mode`, so this change works around [Emacs bug#68600](https://debbugs.gnu.org/cgi/bugreport.cgi?bug=68600).

Also declare `visual-line-mode` value in the :options list, as that value is added to this hook as standard.

### DIFF
--- a/ement-room.el
+++ b/ement-room.el
@@ -2089,10 +2089,10 @@ the previously oldest event."
       ;; We don't really care about the response, I think.
       :then #'ignore)))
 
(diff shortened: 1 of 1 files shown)


phil-s:
Updated the docstring to align with the (prompt) upstream resolution for this bug.

phil-s:
Rebased over current master.

alphapapa:
Thank you.

Issue #251, opened 2024-01-08
User-specified percentage image scaling


phil-s:
I've just tried this and it seems to work.

When using `M-RET` or `<mouse-1>` on an image you can either supply a numeric prefix arg as a scale percentage, or with `C-u` it will prompt you for the percentage.

The `last-prefix-arg` bit for the mouse-driven command was a hunch that seems to be correct, but I've not gone digging to find out exactly why that happens (my *guess* is that it's because there are *two* mouse events when you click-and-release, with `<mouse-1>` following from `<down-mouse-1>`).

```diff
modified   ement-room.el
@@ -4935,14 +4935,19 @@ ement-room-image-scale-mouse
   (let* ((pos (event-start event))
          (window (posn-window pos)))
     (with-selected-window window
-      (ement-room-image-scale (posn-point pos)))))
+      ;; Any prefix arg has been demoted to `last-prefix-arg' by this point.
+      (ement-room-image-scale (posn-point pos) last-prefix-arg))))
 
-(defun ement-room-image-scale (pos)
+(defun ement-room-image-scale (pos &optional scale)
   "Toggle scale of image at POS.
 Scale image to fit within the window's body.  If image is already
 fit to the window, reduce its max-height to 10% of the window's
-height."
-  (interactive "d")
+height.
+
+With a prefix argument, scale the image to the specified
+percentage of its actual size (if numeric), or prompt for a
+percentage (if not numeric)."
+  (interactive "d\nP")
   (pcase-let* ((image (get-text-property pos 'display))
                (window-width (window-body-width nil t))
                (window-height (window-body-height nil t))
@@ -4951,14 +4956,24 @@ ement-room-image-scale
                (new-height (if (= window-height (or (image-property image :max-height) -1))
                                (/ window-height 10)
                              window-height)))
-    (when (fboundp 'imagemagick-types)
-      ;; Only do this when ImageMagick is supported.
-      ;; FIXME: When requiring Emacs 27+, remove this (I guess?).
-      (setf (image-property image :type) 'imagemagick))
-    ;; Set :scale to nil since image scaling commands might have changed it.
-    (setf (image-property image :scale) nil
-          (image-property image :max-width) window-width
-          (image-property image :max-height) new-height)))
+    (if scale
+        (let ((scale (if (consp scale)
+                         (read-number "Scale (% of original): " 100
+                                      'read-number-history)
+                       scale)))
+          (unless (cl-plusp scale)
[...]

PR #255, opened 2024-01-21
Image scaling

phil-s:
Bug fixes and enhancements for image scaling.

### DIFF
--- a/README.org
+++ b/README.org
@@ -324,6 +324,9 @@ Ement.el doesn't support encrypted rooms natively, but it can be used transparen
   - Option ~ement-room-self-insert-chars~ determines which typed characters will start a new message when ~ement-room-self-insert-mode~ is enabled (regardless of whether they are bound to ~self-insert-command~).
   - Option ~ement-room-mode-map-prefix-key~ defines a prefix key for accessing the full ~ement-room-mode-map~ when ~ement-room-self-insert-mode~ is enabled.  (By default this key is ~DEL~.)
 
++ Option ~ement-room-image-margin~ is the number of pixels of margin around image thumbnails.
++ Option ~ement-room-image-relief~ is the number of pixels of shadow rectangle around image thumbnails.
+
 *Changes*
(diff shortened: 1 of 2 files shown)


phil-s:
Currently a draft on account of "Prefix argument prompts for scale" which was under discussion in https://github.com/alphapapa/ement.el/issues/251 (for the moment I've included that change as it was in that previous code, and haven't yet put a lot of thought into alternatives).

phil-s:
I've changed my mind about including #251 in this, partly because `ement-room-image-show` already provides a way of viewing the image at 100%.  I think there are more improvements to be made in this area, but that this particular PR can move forwards without that (and I'm happy with the other commits).

phil-s:
Coincidentally, I just encountered a case where a very wide image was, at its initial/reduced thumbnail height, too wide for the window body.

I could add support for that case in the same manner as the width restrictions for the "maximised" images.

phil-s:
Rebased over current master.

phil-s:
> Coincidentally, I just encountered a case where a very wide image was, at its initial/reduced thumbnail height, too wide for the window body.
>
> I could add support for that case in the same manner as the width restrictions for the "maximised" images.
>
> (This is definitely an edge case though, so it could also be fixed later.)

On review, I'm not sure we have the necessary information at the time it's needed to handle this case as easily.  A backtrace for `ement-room--format-m.image` is:

- ement-room--format-m.image
- ement-room--format-message-body
- ement-room--format-message
- ement-room--format-event
- ement-room--pp-thing
- ewoc--refresh-node
- ewoc--insert-new-node
- ewoc-enter-before
- ewoc-enter-after
- ement-room--insert-event
- ement-room--process-events
- ement-room--buffer
- ement-view-room

I don't think the specific buffer position for the image is available yet.

If I'm right, then I figure that all we could do here is *register* that (for images generally) we wish to do some post-processing after their insertion into the buffer, and run a hook for that (at the end of `ement-room--pp-thing`?) which would enable us to adjust the max-width image property if necessary, based on where the left-hand edge of the image had actually ended up.

(This seems more and more like a "later" thing.)

phil-s:
Rebased over current master.

PR #260, opened 2024-01-27
Fix: Non-obvious "Join it, or load link with browser?" prompt

### LINKED ISSUE #241: Non-obvious "Join it, or load link with browser?" prompt
I selected a chat URL someone had posted, and ement.el prompted me with:

> Room <nil> not joined on current session.  Join it, or load link with browser?

To which my thought was: "Uh, Yes?"

I.e. It's a `completing-read`, but the prompt comes across like a paradoxical `yes-or-yes-p` query.  Guessing what was happening, I typed TAB and got the completions buffer "2 possible completions: Join room, Load link with browser".

I suspect the code was written with an expectation that users are running some completion framework which *automatically* displays the completions, but that's not the default behaviour and I think this isn't the best way to handle such a query.

How about this:?

```el
(let ((read-answer-short t))
  (read-answer
   "Room not joined on current session. (j)oin it, or (l)oad link with browser? "
   '(("join" ?j "Join room in ement.el")
     ("load" ?l "Load link with web browser"))))
```

phil-s:
- Fix #241
- Change: (ement-room-browse-url) Use `read-multiple-choice`.
- Add mouse support (by side-effect).
- Fix: (ement-room-browse-url) Avoid recursive call.
- Add: (ement--read-multiple-choice) Wrapper for `read-multiple-choice` to bypass the hard-coded multi-column formatting in the help buffer (as it tends to wrap poorly).

### DIFF
--- a/README.org
+++ b/README.org
@@ -295,7 +295,8 @@ Ement.el doesn't support encrypted rooms natively, but it can be used transparen
 
 ** 0.15-pre
 
-Nothing new yet.
+*Changes*
++ Improve prompt used when viewing a room that is not joined.  ([[https://github.com/alphapapa/ement.el/issues/241][#241]].  Thanks to [[https://github.com/phil-s][Phil Sainty]].)
 
 ** 0.14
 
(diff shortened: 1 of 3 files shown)


alphapapa:
Thanks!

PR #236, opened 2023-10-31
Compose buffer enhancements

phil-s:
Potentially this can be split into multiple PRs, but as you might expect there would be conflicts in later commits if I pulled out earlier ones, so I didn't want to do that any earlier than necessary.

## What is this?

A collection of changes providing an alternative UI for message composition which *feels* more like the traditional IRC client approach of "just typing your message at the in-buffer prompt", but built on top of Ement.el's pre-existing "compose buffer" feature.

A minimal config for enabling the main features is:
```el
;; Make the message-entry keybindings invoke compose-buffer-centric commands.
(setq ement-room-compose-method 'compose-buffer)
;; Make "just typing" start a new message (see the mode docstring for details).
(ement-room-self-insert-mode 1)
```

The "new" features are all in use in combination as my daily driver, so I encourage anyone who is interested to go ahead and test this.

The following headings summarise the changes.

## Choose minibuffer-centric or compose-buffer-centric behaviour

- User option `ement-room-compose-method` chooses between minibuffer-centric (default) or compose-buffer-centric (new) behaviour.
- Command `ement-room-dispatch-new-message` starts writing a new message using your chosen `ement-room-compose-method`.  (Bound to `RET` in room buffers.)
- Command `ement-room-dispatch-new-message-alt` starts writing a new message using the alternative method.  (Bound to `M-RET` in room buffers.)
- Command `ement-room-dispatch-edit-message` edits a message using your chosen `ement-room-compose-method`.  (Bound to `<insert>` in room buffers.)
- Command `ement-room-dispatch-reply-to-message` replies to a message using your chosen `ement-room-compose-method`.  (Bound to `S-RET` in room buffers.)

## New commands for entering a compose buffer

- `ement-room-compose-edit` edits a message using a compose buffer.
- `ement-room-compose-reply` replies to a message using a compose buffer.

## New commands within a compose buffer

- `ement-room-compose-send-direct` sends a message directly from a compose buffer (without the minibuffer).  (Bound to `C-x C-s` in compose buffers.)
- `ement-room-compose-abort` kills the compose buffer and delete its window.  (Bound to `C-c C-k` in compose buffers.)
- `ement-room-compose-abort-no-history` does the same without adding to `ement-room-message-history`.  (Equivalent to `C-u C-c C-k`.)
- `ement-room-compose-history-prev-message` cycles backwards through `ement-room-message-history`.  (Bound to `M-p` in compose buffers.)
[...]

PR #270, opened 2024-03-18
Improvements and bug fixes for the compose buffer enhancements

phil-s:
A handful of bug fixes and enhancements following on from https://github.com/alphapapa/ement.el/pull/236.


phil-s:
I've cleaned up the earlier WIP commits (just removing the debugging messages and writing proper commit messages), and otherwise this code is what I've been using for the past several days.

I've decided not to worry about the remaining auto-height edge case that I'd seen while testing -- it is harmless, and I believe it is unlikely to affect people in practice in any case, and I didn't wish to complicate the code further on that account.

phil-s:
Force-pushed to squash this addition into the "More robust auto-height behaviour" commit:

```el
          ;; Ask Emacs to "preserve" the new height.  So long as the window
          ;; maintains this height and is displaying this specific buffer, Emacs
          ;; will avoid unnecessary height changes from side-effects of commands
          ;; such as `balance-windows'.  Explicit height changes are allowed.
          ;; We must update this parameter every time we change the height so
          ;; that the "preserved" height value is always correct.
          (window-preserve-size nil nil t)
```

alphapapa:
Thanks, Phil!

PR #271, opened 2024-03-31
Improvements and bug fixes for the compose buffer enhancements (part 2)

phil-s:
A handful of bug fixes and enhancements following on from https://github.com/alphapapa/ement.el/pull/270.

* Use a window-parameter for the compose buffer window auto-height cache
  The old system was usually fine in practice, but this is correct/better.

* Prevent unwanted resizing when cycling through buffers in a window
  This eliminates a potential pain point, whereby the window resizing might easily be unwanted.

* Support an option for deleting compose buffer windows without them being dedicated
  This is a way to get the "old" behaviour back (when the windows were not dedicated by default).

* Eliminate the timer-triggered resize from ement-room-compose-buffer-window-buffer-change-handler
  This no longer seemed necessary, so I took the opportunity to simplify things by removing it.

* Improved code comments.

### DIFF
--- a/ement-room.el
+++ b/ement-room.el
@@ -523,9 +523,15 @@ the option `ement-room-compose-buffer-window-auto-height' is
 enabled (this option generally keeps the windows too small to
 usefully display other buffers).
 
-See also `set-window-dedicated-p'."
+The value `delete' means that windows will not be dedicated, but
+they will still be deleted once the message is sent or aborted
+\(even when they have also been used to display other buffers).
+
+See also `set-window-dedicated-p' and
+`switch-to-buffer-in-dedicated-window'."
   :type '(radio (const :tag "Always" t)
                 (const :tag "Never" nil)
+                (const :tag "Never (but always delete window)" delete)
                 (const :tag "Newly-created windows" created)
                 (const :tag "When auto-height enabled" auto-height)))
 
@@ -4587,6 +4593,7 @@ See also `ement-room-compose-buffer-window-state-change-handler'."
         (when (cl-case ement-room-compose-buffer-window-dedicated
                 (created createdp)
                 (auto-height ement-room-compose-buffer-window-auto-height)
+                (delete nil)
                 (t ement-room-compose-buffer-window-dedicated))
           (set-window-dedicated-p win t))))))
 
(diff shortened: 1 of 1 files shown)


phil-s:
I can't tell whether https://github.com/phil-s/ement.el/commit/14ddcd8d18e8b408893a31478bfde1953aed354d would be pointless overkill, as I've *only* ever used "strongly" dedicated windows (it only just occurred to me today to check the distinction).  I can trivially merge this if you think it's a good idea.  Or make it a regular variable, if we're dubious about anyone ever needing to change it.

alphapapa:
Thanks.  Quick question from one of the comments:

> Perfection would in any case be non-trivial -- consider two compose window side-by-side in a horizontal split, each showing a different compose buffer with a different desired height.  We cannot have the "correct" size for both simultaneously.  The best thing to do would be to maintain the tallest height amongst all conflicting windows at all times -- but that is, again, considerably more complex.

I don't understand that specific example.  I'm envisioning a layout like this:

    +-----------+-----------+
    | room A    | room B    |
    |-----------|-----------|
    | compose A | compose B |
    +-----------+-----------+

In that case, why couldn't the two compose buffers be at the correct size simultaneously?

ement-room.el
```diff
@@ -4449,21 +4453,50 @@ is non-nil."
...
            (cache (if pixelwise
                       (* buflines lineheight)
-                    buflines)))
+                    buflines))
+           (wcache (window-parameter
+                    nil 'ement-room-compose-buffer-window-auto-height-cache)))
       ;; Do nothing if the desired height has not changed.
-      (unless (and (boundp 'ement-room-compose-buffer-window-auto-height-cache)
-                   (eql cache ement-room-compose-buffer-window-auto-height-cache))
+      (unless (and wcache (eql cache wcache))
```
alphapapa, line 4496: An extreme nitpick, probably, but also just a curiosity for me: Why use `eql` instead of `eq` or `equal`?  (AFAIR the only reason to use `eql` is for certain numerical comparisons.)  It might be clearer (to the reader) to use one or the other instead of `eql`.

alphapapa (commented):
Thank you, as always.

ement-room.el
phil-s: The cache valued should always be either an integer or `nil`, so `eql` seemed best (`=` signals an error for `nil`).  It's a bit of added type-checking and also a visual indication that we're dealing with numbers.

alphapapa: Thanks, I learned something today.  :)

alphapapa:
I'm not sure if the problem I described was ever "solved" by any iteration I've tried, so I'd hesitate to merge any code intended to fix it without better understanding the problem.  Especially, if the implementation you have now doesn't use a timer, I'd hesitate to restore one that does.

If you're not seeing this problem, and if no one else is reporting it, let's wait and see; especially, since I probably won't use this option as the default for myself.  If I really need to, I can investigate it myself or file a bug report about it someday.

Thanks for your help.

PR #273, opened 2024-04-05
Docs: Update readme

phil-s:
Per `(info "(org)Escape Character")` I've introduced a zero-width space before the closing paren in "(while writing in minibuffer: ~C-c '~​)" in order to achieve the correct formatting.  This can be visualised using `(set-face-background 'glyphless-char "red")`.

I've substituted` S-<return>` everywhere that `S-RET` appeared, as the latter isn't a valid representation.

### DIFF
--- a/README.org
+++ b/README.org
@@ -190,8 +190,8 @@ Note that if global minor mode ~ement-room-self-insert-mode~ is enabled (by defa
 *Messages*
 
 + Write message: ~RET~
-+ Write reply to event at point (when region is active, only quote marked text): ~S-RET~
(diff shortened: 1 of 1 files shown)


phil-s:
Darn, Github doesn't like what I did with the `C-c '` formatting!

https://github.com/phil-s/ement.el/tree/phil-s/docs?tab=readme-ov-file#room-buffers

I'll see if I can figure out a fix for that...

phil-s:
That was a pain -- I think I now understand why the text had previously been written as `C-c ')` !

Through trial-and-error I've found that a single ZERO WIDTH JOINER after the apostrophe in `C-c '‍` works for both Emacs and Github, so I've pushed that fix.

alphapapa:
Thank you.

PR #278, opened 2024-04-20
Change: (ement-room-mode) Set `filter-buffer-substring-function'

phil-s:
Strip the `line-prefix` and `wrap-prefix` text properties which are used when formatting certain Matrix events, but which should not be copied into other buffers.

This is especially relevant when using `%W` in `ement-room-message-format-spec` to provide indentation when wrapping message body lines.

With this filter configured, we can copy text from an ement room and yank it into another buffer without this formatting being included.  See also `filter-buffer-substring`:

> For example, major or minor modes can use `filter-buffer-substring-function' to exclude text properties that are special to a buffer, and should not be copied into other buffers.


alphapapa:
Thanks, Phil.

PR #293, opened 2024-08-19
Fix: Customize group for ement-room-mode-hook, ement-room-self-insert-mode

phil-s:
These definitions come after defgroup ement-room-prism, so they need an explicit :group argument.


phil-s:
I just spotted both of these in the `ement-room-prism` group.

(My mistake in both cases.)

alphapapa:
Thanks.

PR #294, opened 2024-08-27
Add: global minor mode `ement-sync-watchdog-mode'

### LINKED ISSUE #177: Sometimes auto-synchronisation stops
I don't have much info about this, but sometimes ement simply stops updating rooms on its own.

Today I've been using ement.el for many hours, and I'm running the stand-alone Element client as well (with a different auth token), and I just noticed that Element had a message in a room which ement.el wasn't showing me.  As soon as I typed `g` in that room, ement.el fetched the missing message (which had been sent more than an hour earlier) as well as a bunch of others (my desktop notifications showed me a heap of messages from other rooms in rapid succession).

All I could find in the messages buffer was `ement-room-sync: Wrong type argument: arrayp, nil` which I've noticed happens when typing `g` in the Notifications or Mentions buffers.  I don't know whether that could have made the auto-sync fail?

I've also seen the sync failing overnight, such that I've taken to typing `g` in some room when I start using my laptop in the morning to ensure that the sync is happening.

Maybe some kind of fail-safe system could be triggered periodically to make sure that the sync system gets restarted after a while if necessary?

phil-s:
Global minor mode `ement-sync-watchdog-mode` periodically re-syncs sessions in case of `ement-auto-sync` failures.  Customizing `ement-auto-sync` automatically enables/disables the new mode.

Option `ement-sync-watchdog-interval` determines how frequently the global mode acts (by default, every 5 minutes).

Option `ement-sync-watchdog-sessions` allows you to constrain which sessions are targeted (by default, all sessions).

Closes #177.

### DIFF
--- a/README.org
+++ b/README.org
@@ -299,6 +299,12 @@ Ement.el doesn't support encrypted rooms natively, but it can be used transparen
 
 ** 0.16-pre
 
+*Additions*
+
++ Global minor mode ~ement-sync-watchdog-mode~ periodically re-syncs sessions in case of ~ement-auto-sync~ failures.  Customizing ~ement-auto-sync~ automatically enables/disables the new mode.  (Thanks to [[https://github.com/phil-s][Phil Sainty]].)
+  - Option ~ement-sync-watchdog-interval~ determines how frequently the global mode acts (by default, every 5 minutes).
+  - Option ~ement-sync-watchdog-sessions~ allows you to constrain which sessions are targeted (by default, all sessions).
+
 *Changes*
 
 + Option ~ement-room-coalesce-events~ may now be set to (and defaults to) a maximum number of events to coalesce together.  (This avoids potential performance problems in rare cases.  See [[https://github.com/alphapapa/ement.el/issues/247][#247]].  Thanks to [[https://github.com/viiru-][Arto Jantunen]] for reporting and [[https://github.com/sergiodj][Sergio Durigan Junior]] for testing.)
--- a/ement.el
+++ b/ement.el
@@ -89,6 +89,9 @@
 Hooks are called with one argument, the session that was
 synced.")
 
+(defvar ement-sync-watchdog-timer nil
+  "Timer used by `ement-sync-watchdog-mode'.")
+
 (defvar ement-event-hook
   '(ement-notify ement--process-event ement--put-event)
   "Hook called for events.
@@ -136,7 +139,14 @@ Writes the session file when Emacs is killed."
   :type 'file)
 
 (defcustom ement-auto-sync t
-  "Automatically sync again after syncing."
+  "Automatically sync again after syncing.
+
+Customizing this option also enables or disables
+`ement-sync-watchdog-mode'."
+  :set (lambda (option value)
+         (set-default-toplevel-value option value)
+         (when (fboundp 'ement-sync-watchdog-mode)
+           (ement-sync-watchdog-mode (if value 1 0))))
   :type 'boolean)
 
 (defcustom ement-after-initial-sync-hook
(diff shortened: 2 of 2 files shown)


Icy-Thought:
Feature was tested by connecting to ement -> writing a message -> disconnect from the internet and receive any form of messages while remaining disconnected from the internet -> re-connect to the internet and it auto syncs after the defined time interval.

Very cool feature and thanks for implementing + pushing the change to upstream! LGTM!

alphapapa:
Thanks, Phil.

I'd like to consider simplifying this feature.  For example, I think it would be nice if there were just an option that users could enable, like `ement-reconnect-p`, or something like that, which, when enabled, would automatically run a timer to re-sync a session when a sync times out.  That way it wouldn't require a minor mode, nor a timer that runs at all times.

I realize that that would mean discarding most of this code you've written, and that the code you've written works as-is.

Anyway, what do you think?

phil-s:
If you can see and address specific failure cases like timeouts, I'm all for it.

The thing I rather like about my approach is that the *reason* for the sync failing is irrelevant, and therefore it ought to be pretty robust; whereas "re-sync a session when a sync times out" comes with some assumptions about what it was that went wrong.

I like the idea that even if the failure was due to some edge-case lisp error, something will come along and kick things off again soon enough.

phil-s:
I guess it might be good if the timer-based re-sync notices when everything is fine and does nothing; but conversely, if it can tell that it *needs* to take action, it logs that fact to a file.  That might then provide some better evidence as to whether or not there's any need for the timer.

phil-s:
`ement-sync-watchdog-mode` also has potential use as an *alternative* to `ement-auto-sync` -- a user could leave the latter option off, and let the minor mode re-sync at their preferred interval, if they felt that real-time-ish updates were a bit noisy/distracting, but didn't want it to always have to do it manually.

alphapapa:
> I like the idea that even if the failure was due to some edge-case lisp error, something will come along and kick things off again soon enough.

I understand the appeal of that, but IME that tends to be a bad idea, especially with Matrix's idempotent token-based syncing: if a sync's events trigger an error in Lisp code, retrying the same sync from the same token will usually trigger the same error again next time.  So rather than retrying that sync, potentially infinitely, it's better to stop syncing and show the error so the user can report it and get the bug fixed.  (Note that I haven't seen any errors like that in a long time, so they're generally unlikely; at the same time, with Matrix's gradual evolution, it's probably inevitable that such an error will eventually happen again.)

For the simple cases, like the network being disconnected for a while, fixing that would be as simple as running a timer to retry syncing here: https://github.com/alphapapa/ement.el/blob/3df237176262ecf6def8f0766154fea066009a97/ement.el#L548-L552

phil-s:
That sounds like the next step, then.

How about we keep this PR on the back-burner, and I can follow up down the track if I think it's still helping.

Is there currently a way to check *when* the last sync request for a session was made?  (I suspect the session has a server-generated timestamp, but it's the Emacs clock value I'd want).  I think it would be handy if we tracked that as well, and then anything that's interested can check how long it's been since the last one.  In fact both a "last sync request time" and "last successful sync completion time" per session would be good, I think.

I see a let-bound `sync-start-time (time-to-seconds)` in `ement--sync` but at present it's just used for `message` purposes.

alphapapa:
> That sounds like the next step, then.
> 
> How about we keep this PR on the back-burner, and I can follow up down the track if I think it's still helping.

Sure.  I'll set it as a draft for now.
> 
> Is there currently a way to check _when_ the last sync request for a session was made? (I suspect the session has a server-generated timestamp, but it's the Emacs clock value I'd want). I think it would be handy if we tracked that as well, and then anything that's interested can check how long it's been since the last one. In fact both a "last sync request time" and "last successful sync completion time" per session would be good, I think.
> 
> I see a let-bound `sync-start-time (time-to-seconds)` in `ement--sync` but at present it's just used for `message` purposes.

No, we rely on the curl process to timeout the request accordingly.  AFAIK that is completely reliable.  I've never seen curl fail to either finish receiving a request or give a timeout error.  That's why the retry logic is in the code I linked, which is triggered by curl's returning an error.

PR #297, opened 2024-09-08
Add: (ement-room-ban-user, ement-room-unban-user) New commands

phil-s:
New admin commands for a room's power users.


phil-s:
All done for now.  Tested, and looking good, I think (for an initial version, at least).

phil-s:
Some comments of note from the testing:

> It seemed to work pretty well from this side too: the ban message showed up in the room buffer, and the was moved into the [Left] group. When you unbanned me, the unban event showed up in the room buffer too.

> Rejoining was a bit more work, because I had to get the room ID manually instead of using completion, but it worked once I was invited.

> The other thing we might consider is to implement a convenience feature to also remove all recent events by the banned user, like Element offers to do when you ban someone. Helps with removing spam.
>
> Can always delete the messages manually in the interim, though.

> Interesting side effect: I killed the room buffer and reopened it so I could see and accept the invitation, then after rejoining, the events in the room that I already had are duplicated.  I guess the server thinks we don't already have those events, so it sends them again, but we don't delete them, so we end up with multiple copies. Although I thought we already had code to deduplicate events…

ement-room.el
```diff
@@ -2419,6 +2419,59 @@ these all require at least version 29 of Emacs):
...
+      (if (yes-or-no-p (format "%s user %s <%s> from room %s? "
+                               prompt
+                               (ement-user-displayname user)
+                               (ement-user-id user)
+                               (ement--format-room room)))
+          (list user ement-room ement-session
+                (read-string "Reason (optional): "
+                             nil nil nil 'inherit-input-method))
+        ;; HACK: This isn't really an error, but is there a cleaner way to cancel?
+        (user-error "Aborted")))))
```
alphapapa, line 2441: Looking at, e.g. `isearch-cancel`, it looks like maybe we should use `(signal 'quit nil)` (which I didn't realize before).  What do you think?

ement-room.el
```diff
@@ -2419,6 +2419,59 @@ these all require at least version 29 of Emacs):
...
+
+(defun ement-room--kick-ban-user (type user room session reason successfmt)
+  "Issue the API request for kicking or un/banning a user.
+See `ement-room-kick-user', `ement-room-ban-user', `ement-room-unban-user'.
+TYPE is `kick', `ban', or `unban'."
+  (pcase-let* (((cl-struct ement-user (id user-id)) user)
+               ((cl-struct ement-room (id room-id)) room)
+               (endpoint (format "rooms/%s/%s" room-id type))
+               (content (if (and reason (not (string= "" reason)))
```
alphapapa, line 2450: We could use `string-empty-p` here.

ement-room.el
```diff
@@ -2419,6 +2419,59 @@ these all require at least version 29 of Emacs):
...
+                            (ement-alist "user_id" user-id "reason" reason)
+                          (ement-alist "user_id" user-id ))))
```
alphapapa, line 2452: Nitpick: extra space.  :)

ement-room.el
```diff
@@ -2419,6 +2419,59 @@ these all require at least version 29 of Emacs):
...
+          (list user ement-room ement-session
+                (read-string "Reason (optional): "
+                             nil nil nil 'inherit-input-method))
+        ;; HACK: This isn't really an error, but is there a cleaner way to cancel?
+        (user-error "Aborted")))))
+
+(defun ement-room--kick-ban-user (type user room session reason successfmt)
+  "Issue the API request for kicking or un/banning a user.
+See `ement-room-kick-user', `ement-room-ban-user', `ement-room-unban-user'.
+TYPE is `kick', `ban', or `unban'."
```
alphapapa, line 2446: It would be good to list all the arguments in the docstring, if for no other reason than to silence `checkdoc` in the linting.  :)

ement-room.el
```diff
@@ -2419,6 +2419,59 @@ these all require at least version 29 of Emacs):
...
+
+(defun ement-room-ban-user (user room session &optional reason)
+  "Ban USER from ROOM on SESSION, optionally with REASON."
+  (interactive (ement-room--kick-ban-interactive "Ban"))
+  (ement-room--kick-ban-user 'ban user room session reason
+                             "User <%s> is banned from room %s."))
+
+(defun ement-room-unban-user (user room session &optional reason)
+  "Unban USER from ROOM on SESSION, optionally with REASON."
+  (interactive (ement-room--kick-ban-interactive "Un-ban"))
```
alphapapa, line 2471: Seems like "Un-ban" should be non-hyphenated, for consistency (at least, the API spec doesn't hyphenate it).

ement-room.el
```diff
@@ -2419,6 +2419,59 @@ these all require at least version 29 of Emacs):
...
+                               (ement-user-displayname user)
+                               (ement-user-id user)
+                               (ement--format-room room)))
+          (list user ement-room ement-session
+                (read-string "Reason (optional): "
+                             nil nil nil 'inherit-input-method))
+        ;; HACK: This isn't really an error, but is there a cleaner way to cancel?
+        (user-error "Aborted")))))
+
+(defun ement-room--kick-ban-user (type user room session reason successfmt)
```
alphapapa, line 2443: I think it might be better to pass callback functions (i.e. `(&key then else)`) so we could potentially do other things after the request completes or fails.

Also, maybe the reason should be an optional argument.  What do you think?

alphapapa (changes requested):
Thanks very much for working on this!

ement-room.el
phil-s: Done.

ement-room.el
phil-s: Done.

ement-room.el
phil-s: Fixed.

ement-room.el
phil-s: Done.

ement-room.el
phil-s: I'm unsure about that.

`keyboard-quit` does this:

```el
  (let ((debug-on-quit nil))
    (signal 'quit nil))
```

which seems like what we'd want as well, as we're not triggering this on account of `C-g`.

But then, that's more or less the point of `user-error`:

```el
(defun user-error (format &rest args)
  "Signal a user error, making a message by passing ARGS to `format-message'.
This is like `error' except that a user error (or \"pilot error\") comes
from an incorrect manipulation by the user, not from an actual problem.
In contrast with other errors, user errors normally do not cause
entry to the debugger, even when `debug-on-error' is non-nil.
This can be overridden by `debug-ignored-errors'.
```

My initial reaction is to keep using `user-error`.

ement-room.el
phil-s: I think it's ok for `reason` to be required here, as it's an internal function.

I don't have firm thoughts on using callbacks here.  I don't think it benefits the current functionality, so it could possibly be a refactoring if a use-case arose?

The only thing which springs to mind is your earlier suggestion of "a convenience feature to also remove all recent events by the banned user" which could be a `:then` callback for the `ban` case.

So perhaps it can take optional `:then` and `:else` callbacks, and use the current function as `:then` if it wasn't supplied?

phil-s:
I see it's possible for the prompt to come up as "Ban user nil <@spammer123:matrix.org> ...?", so I should account for them not having a display name.

ement-room.el
```diff
@@ -2419,6 +2419,64 @@ these all require at least version 29 of Emacs):
                                ('remove "removed from"))
                              (ement--format-room space))))))
 
+;;;;; Admin/power commands
+
+(defun ement-room--kick-ban-interactive (prompt)
+  "Interactive arguments for kicking or un/banning a user.
+See `ement-room-kick-user', `ement-room-ban-user', `ement-room-unban-user'.
+Returns (user room session reason)."
+  (ement-room-with-highlighted-event-at (point)
```
alphapapa, line 2428: Taking a fresh look at this, it occurs to me that we need to allow the command to be run in any context, not just when looking at an undesirable message in a room buffer (i.e. as it is now, one would have to locate a message from a user in order to issue a kick/ban, and I can think of various circumstances in which one might want to just select a room and user and execute the action).  Of course, if point is on a message in a room buffer, we ought to offer the room and user as defaults.

ement-room.el
alphapapa: My only hesitation is that pressing C-g isn't really an "incorrect manipulation", so I'm reluctant to use `user-error` and imply that the user did something wrong.  So I think I'd prefer to use `quit`.  (And I should make the same change in other relevant places now that I've realized it.)

ement-room.el
alphapapa: You're right that it doesn't benefit us at the moment, but I think it would be more consistent with the design of similar commands; and as you noted, it would enable further convenience features.  I agree that the callbacks could be optional, i.e. they could have as default values functions which just print the messages as it currently does.

alphapapa (changes requested):
Thanks again for working on this.

alphapapa:
> I see it's possible for the prompt to come up as "Ban user nil <@spammer123:matrix.org> ...?", so I should account for them not having a display name.

We should probably just add a convenience function to format users for display in these cases, e.g.
```el
(cl-defun ement--format-user-id (user &key with-id-p (room ement-room))
  "Return string describing USER.
If ROOM, return the user's displayname in that room, when set;
otherwise, the user's global displayname.  If WITH-ID-P, quote
the displayname and include the user's MXID, like an email
address."
  (let ((displayname (if room
                         ;; FIXME: If a membership state event has not yet been received, this
                         ;; sets the display name in the room to the user ID, and that prevents
                         ;; the display name from being used if the state event arrives later.
                         (ement--user-displayname-in room user)
                       (or (ement-user-displayname user)
                           (ement-user-username user)))))
    (if with-id-p
        (format "%S <%s>" displayname (ement-user-id user))
      displayname)))
```

ement-room.el
phil-s: Agreed.  I had the same thought, but decided it could wait (I think the main difficulty at present is unbanning if you don't have an event from that user handy).  But yeah, this can certainly be improved.

phil-s:
Cross-referencing with https://github.com/alphapapa/ement.el/issues/300

alphapapa:
By the way, I think we should add support for reporting messages to the homeserver admin.  In Element, when I'm deleting spam, my procedure is usually: 1. Report to homeserver admin; 2. Ban user, 3. delete events.

phil-s:
I didn't know that was a thing.

https://spec.matrix.org/latest/client-server-api/#reporting-content

ement-room.el
phil-s: Done.

ement-room.el
phil-s: Done.

phil-s:
Any thoughts on the "score" value when reporting?

A score ranges from -100 (worst/offensive) to zero (best/inoffensive), but I was finding this slightly awkward/ambiguous for UI purposes.

I decided to see what kind of UI Element presented for it, and... it doesn't.  There's no score option at all.  I used it to send a retrospective report for the last spam event, and the HTTP request contained a score of -100 in the JSON.

In the absence of any universally-agreed scale/mapping for the level of offence any given sort of post should warrant, "everything is -100" is certainly a simple solution.

We *could* provide a handful of pre-determined named values, but would need names like "Most offensive" down to "Inoffensive" rather than names like "Spam", as the latter would just be imposing our definitions of how offensive any given thing was.

We could have a very short labelled list like this, and default to... maybe -100, but maybe leaving space for something worse-than-normal?

```
-100 (most offensive)
-75 (very offensive)
-50 (quite offensive)
-25 (slightly offensive)
0 (inoffensive)
```


Thoughts?

alphapapa:
I appreciate your thoughtfulness, but I think we should just do what Element does here.  If the official client doesn't care about the score parameter, then we probably shouldn't bother with it.  (The score parameter was probably one of those things that seemed like a good idea at the time, but as you've aptly shown, doesn't map very well to reality and usability.)

phil-s:
I've gone with "mimic Element by default" (sending -100 automatically without asking the user), but I've left the ability to select a score if a prefix argument is used.  It's a little bit configurable in case anyone really wants to change the default but, on the basis that probably no one will do that, I left that pretty minimal.

alphapapa:
@phil-s Thanks for your work and your patience.  I pushed some "Minor changes" and rebased.  Let me know if you agree that this is ready to merge.

phil-s:
Fair enough ditching the completing-read for score.  I knew that seemed slightly mad by default, but I'd kept it so that users could change the list of options to a short list and then select from a handful of choices.  I'm not fussed about losing that, though.

The refactor has lost the support for reporting *without* a score, though.  Having `-100` as a pre-populated input rather than "what you'll get by default" meant the user could submit an actual empty value, and we could then omit the score from the API request.  With the new code you can't do that (and the documented `nil` behaviour for `ement-room-report-content-score-default` no longer applies, as `read-number` can't cope with that).

I think we need to revert to reading a string in order to support this.

phil-s:
I'd figured everything not labelled **Required** was optional?

alphapapa:
> I'd figured everything not labelled **Required** was optional?

Well, I guess you're right.  :)  Would you mind testing to see what happens if we send a request without the reason and score fields, just to be certain?  (I'd suggest sending it in the testing room we have, and then redacting the reported event; I think the matrix.org homeserver admins will forgive us for one unnecessary report since we're developing a client and the docs are a bit ambiguous about it.)

phil-s:
I reported a test message with no score, and I got a 200 response and the "Content reported." message.

I was tracing the plz call and confirmed it only had `:body "{\"reason\":\"Testing.  Please ignore.\"}"` for the request body.

phil-s:
Heh, ok, need to avoid this:

> Error running timer `plz--respond': (ement-api-error "400: Content must be a JSON object.")

phil-s:
I believe this is good, now.  I'm not seeing any errors from plz in any of the test scenarios.  In each case I get the following from tracing `plz`:

`(plz post "https://matrix-client.matrix.org/_matrix/client/r0/rooms/<roomid>/report/<eventid>?" :headers (("Authorization" . "Bearer <token>") ("Content-Type" . "application/json")) :body "<body>" :body-type text :as json-read :then #f(lambda (_data) [t] (message "Content reported.")) :else ement-api-error :connect-timeout 10 :timeout 60 :noquery t)`

Where `<body>` is one of the following, depending on which optional arguments are supplied:

- `"{}"`
- `"{\"reason\":\"Testing.  Please ignore.\"}"`
- `"{\"score\":0}"`
- `"{\"score\":-10}"`
- `"{\"score\":0,\"reason\":\"Testing.  Please ignore.\"}"`

phil-s:
Untested as yet, but I think we want this as well, so I'll just drop it in here for now.

```el
(defun ement-room-report-delete-ban (event room session reason)
  "Report the current message, delete it, and ban the sender from the room."
  (interactive
   (ement-room-with-highlighted-event-at (point)
     (let ((event (ewoc-data (ewoc-locate ement-ewoc))))
       (unless (ement-event-p event)
         (user-error "No event at point"))
       (if (yes-or-no-p "Report, Delete, and Ban? ")
           (let ((reason (string-trim
                          (read-string "Reason: "
                                       nil ement-room-report-content-reason-history nil
                                       'inherit-input-method))))
             (list event ement-room ement-session reason))
         ;; Aborted.
         (let ((debug-on-quit nil))
           (signal 'quit nil))))))
  ;; Insist on a reason for this particular command.
  (when (string-empty-p reason)
    (user-error "Must supply a reason."))
  ;; Do all the things.
  (let ((user-id (ement-user-id (ement-event-sender event)))
        (room-id (ement-room-id room))
        (score ement-room-report-content-score-default))
    (ement-room-report-content event room session reason score)
    (ement-room-delete-message event room session reason)
    (ement-room-ban-user user-id room-id session reason)))
```

alphapapa:
Thanks, we definitely need some kind of "do all the things" combo command.  I think it'd be best to follow Element's lead here, i.e. when banning a user, maybe a prefix argument or two could be used to indicate that we want to also delete the user's recent content.  So the workflow might be like: 1) report individual events that need reporting; 2) call the "ban user" command with a prefix argument, which would present a confirmation prompt like, "Ban user and delete user's recent events?".  What do you think?

In the long run, it'd be nice to have a way to mark multiple events in a room buffer, and then call a command to act on them (e.g. if not all of a user's recent events should be reported).  I have some ideas about how to facilitate that in the EWOC-based view.  Let me know if you have any thoughts about that.

phil-s:
I hadn't thought about marking messages, but I had pondered iterating over all the user's messages and prompting the admin to report/delete them all.

What about a search loop, highlighting each message from that sender in the room, and asking `y-or-n-p` for whether to add it to the list of messages to deal with?  That might be relatively easy to implement.

alphapapa:
Yeah, I like that idea.  (Not expecting you to code it, necessarily, but feel free to if you are interested.)

phil-s:
Rebased over master, squashing some fixups.

phil-s:
Something like this?

```el
(defun ement-room-select-user-messages (prompt)
  "Collect message events from the sender of the event at point."
  (interactive)
  (if-let* ((node (ewoc-locate ement-ewoc))
            (event (ewoc-data (ewoc-locate ement-ewoc))))
      (ement-room-occur :regexp ""
                        :user-id (ement-user-id (ement-event-sender event)))
    (user-error "No event at point"))
  (cl-assert (derived-mode-p 'ement-room-occur-mode) nil
             "Failed to create an `ement-room-occur-mode' buffer")
  (let (selected)
    (goto-char (point-min))
    (when-let ((node (ewoc-locate ement-ewoc)))
      (while node
        (ewoc-goto-node ement-ewoc node)
        (let ((event (ewoc-data node)))
          (when (and (ement-event-p event)
                     (equal (ement-event-type event) "m.room.message"))
            (ement-room-with-highlighted-event-at (point)
              (when (y-or-n-p (or prompt "Select message?"))
                (push event selected))))
          (setq node (ewoc-next ement-ewoc node)))))
    (quit-window 'kill)
    (nreverse selected)))

(defun ement-room-report-delete-ban-selected (events room session reason)
  "Report and delete selected messages, and ban the sender from the room."
  (interactive
   (progn
     (unless (when-let* ((node (ewoc-locate ement-ewoc))
                         (event (ewoc-data node)))
               (ement-event-p event))
       (user-error "No event at point"))
     (let ((events (ement-room-select-user-messages "Report and delete?")))
       (unless events
         (user-error "No events selected"))
       (if (yes-or-no-p "Report, Delete, and Ban? ")
           (let ((reason (string-trim
                          (read-string "Reason: "
                                       nil ement-room-report-content-reason-history nil
                                       'inherit-input-method))))
             (list events ement-room ement-session reason))
         ;; Aborted.
         (let ((debug-on-quit nil))
           (signal 'quit nil))))))
  ;; Insist on a reason for this particular command.
  (when (string-empty-p reason)
    (user-error "Must supply a reason."))
  ;; Do all the things.
  (let ((user-id (ement-user-id (ement-event-sender (car events))))
        (room-id (ement-room-id room)))
    (ement-room-ban-user user-id room-id session reason)
    (dolist (event events)
      (ement-room-report-content event room session reason -100)
      (ement-room-delete-message event room session reason))))
```

phil-s:
I've pushed a command `ement-room-report-delete-ban-selected` for feedback.

You'll note that at present it doesn't issue the actual report/delete/ban requests (which have been pretty well-tested at this point, so ought to work).  I've commented those lines out from this new function, and instead it just issues a `message` for each action, so that people can play with the UI and see if they like it.

phil-s:
I had an opportunity to use the new command in practice today (after I removed the training wheels, locally), and it worked fine.

I'm still interested in feedback on the UI (including whether it should even be a separate command in the menu, or just rolled into the "ban a user" command).

alphapapa:
Hi @phil-s, I rebased on current master and then pushed https://github.com/alphapapa/ement.el/pull/297/commits/48ebfcd36b340c76ad903bf3e11db752ce3f67cf.  Please let me know what you think about that.  If you concur, then I think we should go ahead and merge this.

> I'm still interested in feedback on the UI (including whether it should even be a separate command in the menu, or just rolled into the "ban a user" command).

I don't have a strong opinion about it yet, but I think the important thing is that, if the user does one action before another (like banning before reporting/redacting content), it should still be easy to do all the things; it shouldn't depend on remembering which command to use first.  Does that make sense?

PR #309, opened 2024-10-12
Add: Custom message format spec for notification buffers

phil-s:
- `ement-notifications-message-format-spec`: New user option.
- `ement-notifications-room-button-separator`: New user option.
- `ement-notifications-log-to-buffer`: Use them.

### DIFF
--- a/README.org
+++ b/README.org
@@ -306,6 +306,7 @@ Ement.el doesn't support encrypted rooms natively, but it can be used transparen
 *Additions*
 
 + When option ~ement-room-images~ is disabled (preventing automatic download and display of images), individual images may be shown by clicking the button in their events.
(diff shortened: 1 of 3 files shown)


phil-s:
There were a couple of different message specs in the code:
```diff
- (ement-room-message-format-spec "%o%O »%W %S> %B%R%t")
- (setq-local ement-room-message-format-spec "[%o%O] %S> %B%R%t"
```

I only observe the first of those formats in the rendered notifications, so maybe the code setting the other can actually be removed?

I've also added a couple of helper code comments like this for locating the event formatters:
```el
;; Formatter definitions: (occur "(\\ement-room-define-event-formatter")
```

I found that a convenience to be able to call with `C-x C-e`, as it's not the easiest thing to locate the code rendering each of those format spec items, and I think newcomers would find it especially useful.  The 'unnecessary' escaping `\\e` is there to prevent those comments themselves being displayed in the results.

PR #310, opened 2024-10-12
Add: Customize groups for Ement faces

phil-s:
Each face now belongs to one of the following new groups:

- `ement-room-faces` (sub-group of `ement-room`)
- `ement-room-list-faces` (sub-group of `ement-room-list`)
- `ement-tabulated-room-list-faces` (sub-group of `ement-tabulated-room-list`)

Additionally, each of those new groups is also a sub-group of another new group `ement-faces` (which is parent to only the three 'faces' sub-groups).

You can therefore navigate to each list of faces from either the customize group the faces used to belong to, or via the new `ement-faces` container.

Also:

- Delete duplicate defgroup ement-room-list.
- Tweak phrasing for defgroup ement-tabulated-room-list for consistency with ement-room-list.

### DIFF
--- a/README.org
+++ b/README.org
@@ -297,6 +297,10 @@ Ement.el doesn't support encrypted rooms natively, but it can be used transparen
 :TOC:      :depth 0
 :END:
 
+** 0.17-pre
+
+Nothing new yet.
+
 ** 0.16
 
 *Compatibility*
--- a/README.org
+++ b/README.org
@@ -299,7 +299,9 @@ Ement.el doesn't support encrypted rooms natively, but it can be used transparen
 
 ** 0.17-pre
 
-Nothing new yet.
+*Additions*
+
++ Customization groups for faces.  (Thanks to [[https://github.com/phil-s][Phil Sainty]].)
 
 ** 0.16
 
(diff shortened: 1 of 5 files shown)


phil-s:
The customize groups are fairly crowded, and there are a *lot* of faces, so I thought it would be beneficial to separate them out into their own sub-groups.

https://www.gnu.org/software/emacs/manual/html_node/elisp/Common-Keywords.html discusses the use of multiple `:group` values.

alphapapa:
Thanks, Phil.

alphapapa:
Thanks, Phil.  I rearranged the new groups to let us avoid having to specify so many faces' groups.  Does this look good to you?  If so, I'll add a changelog entry and merge it.

phil-s:
Unfortunately, moving definitions around is problematic...

```plain
Debugger entered--Lisp error: (error "Invalid face" ement-room-user)
  set-face-attribute(ement-room-user nil :overline nil)
  ement-room-message-format-spec-setter(...)
  custom-initialize-reset(ement-room-message-format-spec ...)
  custom-declare-variable(ement-room-message-format-spec ...)
  ...
  require(ement-room)
```

alphapapa:
Ah, I was afraid of that, but I was too lazy to lint the compilation myself.  :)  I added a changelog entry and pushed your changes as-is.  Thanks.

PR #314, opened 2024-10-13
Wip/fix 243 session saving

### LINKED ISSUE #243: only the last disconnected session is remembered when `(ement-save-sessions)` is set to `t`
### OS/platform

Linux

### Emacs version and provenance

`GNU Emacs 30.0.50 (build 1, x86_64-gentoo-linux-musl, GTK+ Version 3.24.38, cairo version 1.18.0) of 2023-11-26 ` 
it comes from the official gentoo-linux repo; `emerge =app/editors-emacs`

### Emacs command

emacsclient -c

### Emacs frame type

both nox and GUI

### Ement package version and provenance

ement version 0.13; taken via straight.el from github

### Actions taken

I tried to connect to my primary account via `ement-connect`; then I eval'ed `(ement-connect :uri-prefix "http://localhost:8008")` (where 8008 is the port where pantalaion is running on), to connect to my private account;

### Observed results

after calling `ement-disconnect` on both accounts, and re-calling `ement-connect`, it remembers only the last session I disconnected from, e.g. if I use `ement-disconnect RET account1 ement-disconnect RET account2` then `ement-connect` it will only remember account1

### Expected results

I swear that when I had this set up for the first time with two accounts there was a pop-up that let you choose which account to use;  
the expected result is that after calling `ement-connect` it should have a prompt giving you **all** the session you've been logged

### Backtrace

_No response_

### Etc.

I tried on both doom emacs and vanilla emacs, and the result was the same; I can attach the exact same step that I did on both side if it's needed, just let me know if that would be useful.

alphapapa:
@phil-s This is intended to fix https://github.com/alphapapa/ement.el/issues/243.  A few unrelated changes are in here, which I intend to split out before merging.   But to fix the bug required more extensive changes than I...well, I kind of knew it would require a lot of minor changes, which is why I've put it off for so long.

Anyway, it seems to work now, tested in a clean config with multiple accounts.  Would you mind giving the patch a quick visual inspection and letting me know if anything stands out as wrong?  I'm not asking you to test it.  :)

### DIFF
--- a/ement-lib.el
+++ b/ement-lib.el
@@ -167,7 +167,8 @@ include with the request (see Matrix spec)."
   ;; TODO: Document other arguments.
   ;; SPEC: 10.1.1.
   (declare (indent defun))
-  (interactive (list (ement-complete-session)
+  (interactive (list (ement-complete-session :prompt "Make room as: "
+                                             :pred #'ement-session-has-synced-p)
 		     :name (read-string "New room name: ")
 		     :alias (read-string "New room alias (e.g. \"foo\" for \"#foo:matrix.org\"): ")
 		     :topic (read-string "New room topic: ")
@@ -200,7 +201,8 @@ include with the request (see Matrix spec)."
 Then call function THEN with response data.  Optional string
 arguments are NAME, ALIAS, and TOPIC."
   (declare (indent defun))
-  (interactive (list (ement-complete-session)
+  (interactive (list (ement-complete-session :prompt "Make space as: "
+                                             :pred #'ement-session-has-synced-p)
 		     :name (read-string "New space name: ")
 		     :alias (read-string "New space alias (e.g. \"foo\" for \"#foo:matrix.org\"): ")
 		     :topic (read-string "New space topic: ")
@@ -292,7 +294,8 @@ when necessary, and forget the room without prompting."
   "Ignore USER-ID on SESSION.
 If UNIGNORE-P (interactively, with prefix), un-ignore USER."
   (interactive (list (ement-complete-user-id)
-                     (ement-complete-session)
+                     (ement-complete-session :prompt "Ignore user as: "
+                                             :pred #'ement-session-has-synced-p)
                      current-prefix-arg))
   (pcase-let* (((cl-struct ement-session account-data) session)
                ;; TODO: Store session account-data events in an alist keyed on type.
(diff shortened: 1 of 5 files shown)


phil-s:
Can do, but probably not before the end of next week, if that's ok.  I've lisped a bit more than I should have today already, and the next few days aren't going to be an option.

ement-lib.el
```diff
@@ -772,14 +777,18 @@ THEN and ELSE are passed to `ement-api', which see."
...
-    (_ (let* ((ids (mapcar #'car ement-sessions))
-              (selected-id (completing-read prompt ids nil t)))
-         (alist-get selected-id ement-sessions nil nil #'equal)))))
+    (0 (user-error "No sessions.  Call `ement-connect' to log in"))
+    (_ (let ((sessions (cl-remove-if-not pred ement-sessions :key #'cdr)))
+         (pcase (length sessions)
+           (1 (cdar sessions))
+           (_ (let* ((ids (mapcar #'car sessions))
+                     (selected-id (completing-read prompt ids nil t)))
+                (alist-get selected-id sessions nil nil #'equal))))))))
```
phil-s, line 791: I think a lot of the `ement-complete-session` uses could reasonably be expected to act on the session for the current room buffer.  How about we make that call to `completing-read` specify the id for `ement-session` as its `DEF`ault argument?

We'd then want to make the prompt `(format "%s [%s]: " prompt id)` in the case when there was a default.

phil-s: I was also wondering whether there could be cases in practice where the user thought they were in multiple sessions, but in fact only one of them was connected, and the "don't prompt when only one session is connected" behaviour caused them to do something in the wrong session.

It might be reasonable to have a user option to say "prompt whenever there are multiple sessions, even if only one of them is connected".

phil-s (commented):
Apologies for not getting around to this sooner!

ement.el
```diff
@@ -134,8 +131,14 @@ Writes the session file when Emacs is killed."
              (add-hook 'kill-emacs-hook #'ement--kill-emacs-hook)
            (remove-hook 'kill-emacs-hook #'ement--kill-emacs-hook))))
 
-(defcustom ement-sessions-file "~/.cache/ement.el"
-  ;; FIXME: Expand correct XDG cache directory (new in Emacs 27).
+(defcustom ement-sessions-file
+  (or (cl-loop for filename in
+               (list "~/.cache/ement.el"
```
Thaodan, line 136: The order here is wrong. XDG says it should be first xdg-cache. In any case the fallback is wrong as the xdg-cache-home function falls back to `~/.cache` on it's own.

Thaodan:
Using the PR for about a little over a  week now.
It does work quite well.
Just initially you have to redo any session who came after the first since it wasn't saved of course.

PR #322, opened 2025-04-06
Add: (ement-room-hide-redacted-message-content) New option

alphapapa:
@phil-s What do you think about this as a new default?  

I had considered implementing a button to toggle the visibility of redacted messages, but I'm not sure the complexity would be worth it.  This seems like a good and simple default behavior to protect users from seeing unpleasant content that has since been redacted, while still allowing it to be overridden if necessary.


phil-s:
Yes, I think that's a good idea given recent events.

I've tested redacting both text and image messages in Element to check that they're automatically replaced with the "redacted" message in Ement, and it works well as far as I can see.

I don't see any problem with using the simplest approach.  The user option is the important thing, and the implementation could always be reworked later.

PR #323, opened 2025-04-22
Support downloading media

viiru-:
Add a function for downloading media, and use it as the default action for m.audio, m.file and m.video.

Now that authenticated media is the expected situation opening these as links in an external browser no longer works, so try this instead.

"Open in the chosen application for this mime type" would be nice as well, but based on looking at the way Gnus does this it seems harder than expected. Perhaps there is something I don't understand here..

### DIFF
--- a/README.org
+++ b/README.org
@@ -301,6 +301,7 @@ Ement.el doesn't support encrypted rooms natively, but it can be used transparen
 
 *Additions*
 
++ Command ~ement-room-download-file~, which downloads the file in the event at point (for image, audio, video, and file messages).  ([[https://github.com/alphapapa/ement.el/pull/323][#323]].  Thanks to [[https://github.com/viiru-][Arto Jantunen]].)
 + Customization groups for faces.  (Thanks to [[https://github.com/phil-s][Phil Sainty]].)
 + Option ~ement-room-hide-redacted-message-content~, which hides the content of redacted messages by default.  It may be disabled to keep redacted content visible with a strikethrough face, which may be useful for room moderators, but users should keep in mind that doing so will leave unpleasant content visible in the current session, even after being redacted by moderators.
 + Option ~ement-room-list-avatars-generation~: if disabled, SVG-based room avatars are not generated.  This option automatically tests whether SVG support is available in Emacs, and should allow use with builds of Emacs that lack =librsvg= support. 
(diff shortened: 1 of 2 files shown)


alphapapa:
Thanks, that looks useful.  I pushed a few changes on top of your branch.  If you concur, I'll merge it.

alphapapa:
Thanks for adding the bindings.  Part of me feels like it should be an uppercase letter, in case we want to use lowercase-D for a more mundane command in the future, but I don't feel too strongly about it.  What do you think?

viiru-:
> Thanks for adding the bindings. Part of me feels like it should be an uppercase letter, in case we want to use lowercase-D for a more mundane command in the future, but I don't feel too strongly about it. What do you think?

Either works fine for me, all I need is the command somewhere so that I get access to the flies sent to me. I'll switch it to uppercase, for both the room command and the transient, I guess?

alphapapa:
Thanks, @viiru-!

phil-s:
The downloading fails for me for the video posted by amadaluzia in the Emacs room on 2025-04-16/17 (depending on your time zone).

```
((:id . "$ir0NiwBgEwMXIqsYvnbz0pADpE2gPTU43lgi5hQMN9s")
 (:sender . "@amadaluzia:tchncs.de")
 (:content (body . "16_Apr_2025-20:17:04.mp4")
	   (info (duration . 16751) (h . 1080)
		 (mimetype . "video/mp4") (size . 5227651)
		 (thumbnail_info (h . 337) (mimetype . "image/webp")
				 (size . 15236) (w . 600))
		 (thumbnail_url
		  . "mxc://tchncs.de/1dfc70b7f402d3af38c2f7c4b7bc5955bd1bf85b1912586167333683200")
		 (w . 1920))
	   (msgtype . "m.video")
	   (url
	    . "mxc://tchncs.de/6b7f989f5ca789c3d4169756eff8a59a2571a6401912586226855051264"))
 (:origin-server-ts . 1744831067529) (:type . "m.room.message")
 (:state-key) (:unsigned (membership . "join")) (:receipts) (:local))
```

I get `ement-room-download-file: File already exists: "/path/to/Downloads/"` when I use `D`, and if I click the mouse I get `button-activate: Wrong number of arguments: #<subr ement-room-download-file>, 1`

phil-s:
And re-testing with a fresh config, I get `error: (void-variable eww-download-directory)` from `D`, so I assume it's also missing a `(require 'eww)`.

alphapapa:
Thanks for testing and reporting, @phil-s.  The two commits I just pushed should solve those things, with one exception: pressing `C-u` before clicking a download link doesn't prompt for a filename the same way `C-u D` does.  I don't know how to fix that right now, and I'm going to leave that for later.  :)

phil-s:
Confirming that the fixes work for me.

PR #335, opened 2025-11-10
Fix: (ement-room.el) Invalid JSON type

Zzull:
The JSON sent to cancel the typing notification was rejected by the server:

```
Error running timer ‘plz--respond’: (ement-api-error "400: M_BAD_JSON:
deserialization failed: invalid type: null, expected a boolean at line
1 column 30")
```

### DIFF
--- a/ement-room.el
+++ b/ement-room.el
@@ -2640,7 +2640,9 @@ the previously oldest event."
                (endpoint (format "rooms/%s/typing/%s"
                                  (url-hexify-string room-id) (url-hexify-string user-id)))
                (data (ement-alist "typing" typing "timeout" 20000)))
-    (ement-api session endpoint :method 'put :data (json-encode data)
+    (ement-api session endpoint :method 'put
+      :data (let ((json-false nil))
+              (json-encode data))
       ;; We don't really care about the response, I think.
       :then #'ignore)))
 


alphapapa:
Hello,

I've never seen this problem before.  What server does your homeserver run?

Zzull:
> I've never seen this problem before. What server does your homeserver run?

Hello! My homeserver is running continuwuity.

alphapapa:
I guessed you might say that.  That server seems to be a bit restrictive in what it accepts.  This isn't the first time I've heard of such issues with it.  I'm generally in favor of strictly specified APIs, but Postel's Law has been a guiding principle of the Internet for many decades, so I'm not sure this is a good thing for that server to insist upon.  If the canonical Matrix servers (Synapse and Dendrite) accept what we send...

Anyway, I don't mind "fixing" this, but this isn't the best way to do it, because this "pollutes" a Lisp API with specifics of JSON serialization, which shouldn't matter here.  I think we'd probably want to do something like this:

```diff
diff --git a/ement-room.el b/ement-room.el
index a794b78..1ba484d 100644
--- a/ement-room.el
+++ b/ement-room.el
@@ -2640,7 +2640,9 @@ (cl-defun ement-room--send-typing (session room &key (typing t))
                (endpoint (format "rooms/%s/typing/%s"
                                  (url-hexify-string room-id) (url-hexify-string user-id)))
                (data (ement-alist "typing" typing "timeout" 20000)))
-    (ement-api session endpoint :method 'put :data (json-encode data)
+    (ement-api session endpoint :method 'put
+      :data (let ((json-false nil))
+              (json-encode data))
       ;; We don't really care about the response, I think.
       :then #'ignore)))
```

That seems to produce the desired effect, e.g.
```el
(let ((json-false nil))
  (json-encode '(("typing" . nil))))

"{\"typing\":false}"
```
What do you think?  Thanks.

Zzull:
It's indeed cleaner. I tried it and confirm that continuwuity isn't throwing tantrums anymore.

PR #338, opened 2025-12-04
fix: Make sure the matched string is not simply a substring of a bigger word

Konubinix:
If one displayname is 'sam' and another's nickname is 'samy', then all events to samy were recognized as mentionning sam as well.

See the issue https://github.com/alphapapa/ement.el/issues/240

### DIFF
--- a/ement-room.el
+++ b/ement-room.el
@@ -4250,7 +4250,8 @@ HTML is rendered to Emacs text using `shr-insert-document'."
     (when body
       (cl-macrolet ((matches-body-p
                       (form) `(when-let ((string ,form))
-                                (string-match-p (regexp-quote string) body))))
+                                (string-match-p (rx word-start (literal string) word-end)
+                                                body))))
         (or (matches-body-p (ement-user-username user))
             (matches-body-p (ement--user-displayname-in room user))
             (matches-body-p (ement-user-id user)))))))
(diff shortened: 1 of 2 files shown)


alphapapa:
Thanks. I generally prefer to use the `rx` macro for constructing regexps. If you're not familiar with it, the docstring explains it thoroughly, or I can make that change. 

Something that might be really helpful would be to think about scenarios in which this change might still not behave as expected or as intended, and write a few examples. That could help guide future development, and potentially save time by avoiding the retracing of steps. Or in other words, maybe we should write out a few test cases, even in plain English. 

Thanks for your help with this.

Konubinix:
I used rx and added some test cases. I added a few one commented because I would have loved the code to deal with them properly but I could not find a regex that would work. Using the word-start and word-end is still a good way to get rid of a lot of false positives.

phil-s:
`(rx (regexp (regexp-quote foo)))` can just be `(rx (literal foo))`

Konubinix:
@phil-s thank you for the review. I took it into account in the code.

phil-s:
You might break the line here for formatting, but otherwise this PR looks good to me.

```diff
- (string-match-p (rx word-start (literal string) word-end) body))))
+ (string-match-p (rx word-start (literal string) word-end)
+                 body))))
```

Konubinix:
phil-s ***@***.***> writes:

> phil-s left a comment (alphapapa/ement.el#338)
>
> You might break the line here for formatting, but otherwise this PR looks good to me.
>
> ```diff
> - (string-match-p (rx word-start (literal string) word-end) body))))
> + (string-match-p (rx word-start (literal string) word-end)
> +                 body))))
> ```

Done

-- 
Konubinix
GPG Key    : 7439106A
Fingerprint: 5993 BE7A DA65 E2D9 06CE  5C36 75D2 3CED 7439 106A

PR #353, opened 2026-06-24
Fix: (ement-notify--room-unread-p) Handle absent unread counts

bhw-foss:
`ement-notify--room-unread-p` reports a room as unread whenever its
`unread-notifications` slot is absent (nil), because `(equal 0 nil)`
returns nil.  This is the case for rooms muted on the server: their
notification counts stay at 0 and are therefore omitted from `/sync`,
so the slot is never populated.  The result is that every message in a
muted room is still logged to the notifications buffer.

Coerce absent counts to 0 so such rooms are treated as read; rooms
with a genuine positive count are unaffected.

Thanks for Ement! It's great to use.


Thaodan:
You can't submit PR's for GNU projects which have been created with the help of AI.
You have to be able to own the copyright, and submit yours to the FSF.

PR #354, opened 2026-07-31
Fix: (ement--update-room-buffers) Clear event queues of bufferless rooms

bhw-foss:
Ephemeral events (m.typing/m.receipt) are pushed to every room's
ephemeral slot on every sync, and new timeline/account-data events are
queued in every room's local slot, but these queues were only cleared
for rooms having a live buffer.  For every other room they accumulated
for the life of the session (noticeable, e.g., with bridged accounts
whose rooms produce many receipts).

The queues only matter for updating existing room buffers: when a room
buffer is created, it draws events from the room's timeline and state
slots (and already discards new-events, see ement-room--buffer), and
ephemeral events are already processed globally in
ement--push-joined-room-events.  So clear the queues for all rooms on
every sync, not only for rooms with buffers.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>


Thaodan:
Could open an issue on this? I like the idea but I'm not sure on the change/how to handle.

Also again there's the whole AI issue.

bhw-foss:
I'll mark it as a draft but feel free to close. I respect the licensing that comes with ement.el being in GNU ELPA so no AI use. I figured since I (ahem, claude) solved this in my own config by abusing "advice" I might as well document it for the next traveler down this road. 

One other thing I find immensely useful is having the AI add a "search all ement rooms with this regexp" to ement.el.

